Issue #6, opened 2020-01-20
secrets


dweomer:
**Is your feature request related to a problem? Please describe.**
Unable to inject any information into the upgrade pod/container other than what is on the host/node filesystem. We would like to inject secrets.

**Describe the solution you'd like**
The **_PlanSpec_** should have a way to specify name and mount-path for secrets available to the apply job/pod.

**Describe alternatives you've considered**
Leveraging **_PodPresets_** was considered but they are applied passively. Ideally we would like the update of the content of a secret to trigger application of the upgrade job.

PR #10, opened 2020-01-28
secrets support

dweomer:
- Add `secrets` references to **Plan** spec. Changes in referenced `secrets` can trigger re-application of a **Plans**.
- The label that a **Plan** places on an applied `node` is no longer the `status.latestVersion` but is instead a sha224 digest of `status.latestVersion` and the key-order-value(s) of each secret's `data` in the order each `secret` is specified. This value is represented as `status.latestHash`

Addresses #6


pkg/upgrade/job/job.go
```diff
@@ -182,6 +183,9 @@ func NewUpgradeJob(plan *upgradeapiv1.Plan, serviceAccountName, nodeName, contro
 						Env: []corev1.EnvVar{{
 							Name:  "SYSTEM_UPGRADE_PLAN_NAME",
 							Value: plan.Name,
+						}, {
+							Name:  "SYSTEM_UPGRADE_PLAN_LATEST_HASH",
```
galal-hussein: Why do you inject the hash as env variable into the the pod?

pkg/upgrade/plan/plan.go
```diff
@@ -94,6 +97,24 @@ func RegisterHandlers(ctx context.Context, serviceAccountName, controllerNamespa
 		return obj, nil
 	})
 
+	// secret events referred to by a plan (potentially) trigger that plan
+	secrets.OnChange(ctx, controllerName, func(key string, obj *corev1.Secret) (*corev1.Secret, error) {
+		if planList, err := plans.Cache().List(controllerNamespace, labels.Everything()); err != nil {
+			logrus.Error(err)
```
galal-hussein: shouldn't this return nil, err

pkg/upgrade/job/job.go
dweomer: It's general purpose and something might need it. 🤷‍♂ 

The latest version is/was there so I decided to replicate that behavior with the hash.

pkg/upgrade/plan/plan.go
dweomer: That is a good question. I did intentionally write this secrets-trigger-plan logic to make a best effort attempt and I figured if listing secrets is causing problems it is probably a permissions problem and therefore not something likely to resolve itself (thus the re-enque implied by returning an error seemed gratuitous). I am open to changing this.

galal-hussein: I am just wondering what will be the result of returning nil object for both the secret and the error to the controller when a listing error happen?

dweomer: It should have fell through to the last line of the function that was returning `obj, nil`. I ran your comment by @ibuildthecloud and he agreed so I changed both the secret and node OnChange events to return the potential error when listing plans to (potentially) trigger

PR #12, opened 2020-01-30
uncordon on upgrade job completion

### LINKED ISSUE #9: uncordon after successful upgrade
After a job has completed successfully it is left to a human operator to uncordon an upgraded node (if it was cordoned or drained).

After a plan has been successfully applied to a node with `cordon`/`drain` configured the controller should uncordon the node.

n/a

**Additional context**
n/a

dweomer:
Sets `node.spec.unschedulable` to `false` after a job completes successfully.
This happens in the same operation that sets the **Plan**-specific tag on job success.

Fixes #9

### DIFF
--- a/pkg/upgrade/plan/plan.go
+++ b/pkg/upgrade/plan/plan.go
@@ -72,6 +72,9 @@ func RegisterHandlers(ctx context.Context, serviceAccountName, controllerNamespa
 	nodes := coreFactory.Core().V1().Node()
 	secrets := coreFactory.Core().V1().Secret()
 
+	nodeCache := nodes.Cache()
+	planCache := plans.Cache()
+
 	// cluster id hack: see https://groups.google.com/forum/#!msg/kubernetes-sig-architecture/mVGobfD4TpY/nkdbkX1iBwAJ
 	systemNS, err := coreFactory.Core().V1().Namespace().Get(metav1.NamespaceSystem, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/upgrade/plan/plan.go
```diff
@@ -132,6 +131,9 @@ func RegisterHandlers(ctx context.Context, serviceAccountName, controllerNamespa
 								return obj, err
 							}
 							node.Labels[planLabel] = version
+							if node.Spec.Unschedulable {
```
dweomer: Should we mark the node as schedulable no matter what (as we are doing here) or should we only flip it back if cordschedulableon is true or drain is not nil?

pkg/upgrade/plan/plan.go
```diff
@@ -125,13 +124,20 @@ func RegisterHandlers(ctx context.Context, serviceAccountName, controllerNamespa
 				defer plans.Enqueue(obj.Namespace, planName)
 				if obj.Status.Succeeded == 1 {
 					planLabel := upgradeapi.LabelPlanName(planName)
-					if version, ok := obj.Labels[planLabel]; ok {
+					if planHash, ok := obj.Labels[planLabel]; ok {
 						if nodeName, ok := obj.Labels[upgradeapi.LabelNode]; ok {
 							node, err := nodes.Get(nodeName, metav1.GetOptions{})
```
galal-hussein: just a nit: shouldn't we use a node cache to be quicker on a large scale here?

dweomer: @galal-hussein plan cache too then I would think.

dweomer: pushed new commit that uses caches

PR #16, opened 2020-01-30
per-plan service account name

### LINKED ISSUE #7: service account for upgrades
Right now the service account that the controller runs with (having `cluster-admin` role binding) is passed off to the pods that run upgrades. This is far to much permission and should be paired down to just enough to run the pods (at a minimum, whatever `kubectl drain` needs).

Ability to specify a service account name for the upgrade job/pod to run with.

n/a

**Additional context**
n/a

dweomer:
The controller's service account is no longer utilized for upgrade jobs. This means that, unless specified on the **Plan**, upgrade jobs will get the `default` service account for the namespace.

Fixes #7 

P.S. Also lumped in here are some build script changes.

### DIFF
--- a/main.go
+++ b/main.go
@@ -91,7 +91,7 @@ func Run(c *cli.Context) {
 		logrus.Fatal(err)
 	}
 	ctx := signals.SetupSignalHandler(context.Background())
-	if err := upgrade.StartController(ctx, cfg, threads, namespace, serviceAccountName, name); err != nil {
+	if err := upgrade.StartController(ctx, cfg, threads, namespace, name); err != nil {
 		logrus.Fatalf("Error starting: %v", err)
 	}
 	<-ctx.Done()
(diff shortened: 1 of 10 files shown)


galal-hussein:
quick question: does the PR consider the RBAC role given for this plan? or is it up to the user to give permission to plan.spec.serviceaccount to change cordon/uncordon the nodes, if yes then should we add a specific permission to the job's service account into the manifest directory?

dweomer:
> quick question: does the PR consider the RBAC role given for this plan? or is it up to the user to give permission to plan.spec.serviceaccount to change cordon/uncordon the nodes, if yes then should we add a specific permission to the job's service account into the manifest directory?

I left that as an exercise for the user because, provided the upgrade does not perform cordon/drain, the default service account should be adequate.

galal-hussein:
@dweomer got it, thanks for explaining

PR #17, opened 2020-01-31
init container: prepare

### LINKED ISSUE #8: prepare prior to cordon/drain+upgrade
Would like to (optionally) run a process that must exit successfully before invoking the (optional) cordon/drain and then moving on to the `upgrade` process. 

Add `prepare` "container spec" that looks a lot like the current `upgrade` spec.

n/a

**Additional context**
n/a

dweomer:
The `prepare` container has the same config as the `upgrade` container.
If configured it will be the first container to run in the pod. This enables
the use-case where a plan to be applied to worker nodes may want to wait on
another plan being applied to master nodes before incurring drain/cordon on
the workers. There are likely many other legitimate use-cases.

Fixes #8


dweomer:
Also fixed a bug where the values for `drain`'s `--delete-local-data` and `--ignore-daemonsets` were transposed.

pkg/upgrade/job/job.go
```diff
@@ -138,116 +135,46 @@ func NewUpgradeJob(plan *upgradeapiv1.Plan, nodeName, controllerName string) *ba
...
 	if drain != nil {
 		args := []string{"drain", nodeName, "--pod-selector", `!` + upgradeapi.LabelController}
 		if drain.IgnoreDaemonSets == nil || *plan.Spec.Drain.IgnoreDaemonSets {
-			args = append(args, "--delete-local-data")
+			args = append(args, "--ignore-daemonsets")
 		}
 		if drain.DeleteLocalData == nil || *drain.DeleteLocalData {
-			args = append(args, "--ignore-daemonsets")
+			args = append(args, "--delete-local-data")
 		}
```
dweomer: This is why the next milestone will see me writing the sonobuoy plugin: to catch this kind of derp early.

pkg/upgrade/job/container.go
```diff
@@ -0,0 +1,102 @@
...
+	return func(container *corev1.Container) {
+		container.SecurityContext = securityContext
+	}
+}
+
+func withImageTag(tag string) containerOpt {
+	return func(container *corev1.Container) {
+		image := container.Image
+		if p := strings.Split(image, `:`); len(p) > 1 {
+			image = strings.Join(p[0:len(p)-1], `:`)
```
galal-hussein: I am wondering if the code also check if the user specified the tag in the spec.image for example:
```
Spec:
  Image: rancher/k3s-upgrade:v1.17.2
```
correct me if i am wrong but according to the code it will ignore the v1.17.2 value if the version is not specified

dweomer: @galal-hussein  yes, this will ignore a tag on the image if supplied for the `upgrade` container because the tag for this image should be the resolved version aka `status.latestVersion`

Issue #11, opened 2020-01-29
sonobuoy plugin for tests


dweomer:
- [x] write tests
- [x] publish image

dweomer:
Addressed by #26

PR #20, opened 2020-02-04
testing: unit tests leveraging ginkgo and gomega

dweomer:
- Refactored most of the container creation stuff into the `container` package which is where all the coverage is for now.
- Added `upgrade.cattle.io/version` label to Jobs and their Pods to complement existing `upgrade.cattle.io/plan` and `upgrade.cattle.io/node` (Versions used to be on `plan.upgrade.cattle.io/<plan-name>` but that is a hash/digest now)


dweomer:
This is in preparation for #11

pkg/upgrade/container/container_test.go
```diff
@@ -0,0 +1,150 @@
...
+		zeroContainer *corev1.Container
+		testContainer corev1.Container
+		testOption    container.Option
+	)
+	BeforeEach(func() {
+		zeroContainer = &corev1.Container{}
+		testContainer = *zeroContainer
+		Expect(zeroContainer).To(Equal(&testContainer))
+		Expect(testContainer).To(BeZero())
+	})
```
cbron: Guessing you don't need to teardown ?

pkg/upgrade/upgrade_suite_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+)
+
+func TestUpgrade(t *testing.T) {
+	gomega.RegisterFailHandler(ginkgo.Fail)
+	ginkgo.RunSpecs(t, "Upgrade Suite")
+}
+
+// Declarations for Ginkgo DSL
```
cbron: Are all these auto-generated ? Or can we pull them in dynamically or something. I'm not familiar enough with ginkgo.

Makefile
```diff
@@ -10,6 +10,9 @@ TARGETS := $(shell ls scripts)
 $(TARGETS): .dapper
 	./.dapper $@
 
+e2e:
+	./scripts/verify
```
cbron: This doesn't exist yet ?

pkg/upgrade/container/container_test.go
```diff
@@ -0,0 +1,150 @@
+package container_test
```
cbron: When you're settled on a test framework, would be good to add a "how to run test suite" section to docs.

pkg/upgrade/container/container_test.go
dweomer: In the cases of these tests, no, I don't. This is because the `BeforeEach` re-initializes before every test case.

pkg/upgrade/upgrade_suite_test.go
dweomer: See: https://onsi.github.io/ginkgo/#avoiding-dot-imports
```
# *_suite_test.go one-time generated via:
ginkgo bootstrap --nodot
# with some minor cleanup/edits 
```

Makefile
dweomer: pre-cruft

PR #26, opened 2020-02-08
testing: sonobuoy plugin for e2e testing

dweomer:
- includes removal of controller serviceAccountName references and envvar
- image tarballs a la `docker save` are added to dist/artifacts
- sonobuoy plugin yaml added to dist/artifacts
- introduces e2e-tests tag suffix image that can be used for the sonobuoy plugin


e2e/suite/upgrade_test.go
```diff
@@ -0,0 +1,67 @@
+package suite_test
+
+import (
+	"time"
+
+	"github.com/rancher/system-upgrade-controller/e2e/framework"
+	upgradeapiv1 "github.com/rancher/system-upgrade-controller/pkg/apis/upgrade.cattle.io/v1"
+)
+
+var _ = Describe("Upgrade", func() {
```
dweomer: everything else in this commit exists to run these three simple tests. will add more once this lands on master

e2e/suite/upgrade_test.go
```diff
@@ -0,0 +1,67 @@
...
+			plan, err = e2e.CreatePlan(plan)
+			Expect(err).ToNot(HaveOccurred())
+
+			plan, err = e2e.PollPlanCondition(plan.Name, upgradeapiv1.PlanLatestResolved, 2*time.Second, 30*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+		})
+		It("should not resolve", func() {
+			Expect(upgradeapiv1.PlanLatestResolved.MatchesError(plan, "Error", upgradeapiv1.ErrPlanUnresolvable)).To(BeTrue())
+		})
+	})
```
dweomer: test for unresolavable plan

e2e/suite/upgrade_test.go
```diff
@@ -0,0 +1,67 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			plan, err = e2e.PollPlanCondition(plan.Name, upgradeapiv1.PlanLatestResolved, 2*time.Second, 30*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+		})
+		It("should resolve", func() {
+			Expect(upgradeapiv1.PlanLatestResolved.IsTrue(plan)).To(BeTrue())
+			Expect(upgradeapiv1.PlanLatestResolved.GetReason(plan)).To(Equal("Version"))
+		})
+	})
```
dweomer: assert that a plan with a version resolves

e2e/suite/upgrade_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+			plan, err = e2e.PollPlanCondition(plan.Name, upgradeapiv1.PlanLatestResolved, 2*time.Second, 30*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+		})
+		It("should resolve", func() {
+			Expect(upgradeapiv1.PlanLatestResolved.IsTrue(plan)).To(BeTrue())
+			Expect(upgradeapiv1.PlanLatestResolved.GetReason(plan)).To(Equal("Channel"))
+		})
+	})
+})
```
dweomer: assert that a plan with a channel resolves

e2e/cluster/local/docker-compose.yaml
```diff
@@ -0,0 +1,41 @@
...
+      context: ./images/k3s/.
+      args:
+        - ARCH=${ARCH}
+    command: server --no-deploy=traefik,metrics-server
+    environment:
+      - K3S_TOKEN=e2e
+    hostname: local-leader
+    privileged: true
+    ports:
+      - "172.17.0.1:6443:6443" # k3s
```
galal-hussein: Will the address be always 172.17.0.1 for the host (dapper container i assume?)

e2e/cluster/local/docker-compose.yaml
```diff
@@ -0,0 +1,41 @@
...
+services:
+
+  leader:
+    container_name: local-leader
+    image: "e2e/cluster/local/k3s"
+    build:
+      context: ./images/k3s/.
+      args:
+        - ARCH=${ARCH}
+    command: server --no-deploy=traefik,metrics-server
```
galal-hussein: Can we use disable here instead of no-deploy

e2e/cluster/local/docker-compose.yaml
```diff
@@ -0,0 +1,41 @@
+version: "3.7"
+
+volumes:
+  kubeconfig:
+    name: e2e-local-kubeconfig
+
+services:
+
+  leader:
```
galal-hussein: nit: I think server/agent nomenclature is more accurate than leader/worker

e2e/cluster/local/docker-compose.yaml
dweomer: It's the default for docker0. This worked for me on a linux box, my mac, and it passed drone so that's good enough for me for now. I do agree that this should be more flexible in the future but I just wanted to get it off my plate.

e2e/cluster/local/images/k3s/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Install the all-in-one binary so we can copy our run-time images into the image
```
galal-hussein: Is there a reason you are not using the official Image directly in the docker-compose?

e2e/cluster/local/scripts/cluster-up
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e
```
galal-hussein: nit: maybe a `DEBUG` variable can be useful here to give more visibility with -x for example

.drone.yml
```diff
@@ -215,6 +256,27 @@ steps:
...
+- name: e2e-tests
+  image: plugins/manifest:1.0.2
+  settings:
+    username:
+      from_secret: docker_username
+    password:
+      from_secret: docker_password
+    platforms:
+      - linux/amd64
+      - linux/arm64
```
galal-hussein: Is there a reason arm is not included here?

dweomer: there is currently no sonobuoy client built for arm

e2e/cluster/local/images/k3s/Dockerfile
dweomer: Yes, I tried to capture in this comment. I have to recreate a k3s image because the official image declares `/var/lib/rancher` as a VOLUME which means subsequent additions to that area will never show up in an image. If, however, in a Dockerfile you copy content into a location and then declare it a VOLUME afterwards and then run the image with an non-bind mount volume at that location, docker will by default at first-run copy the contents from the image into the volume. This is the trick that I am leveraging.

e2e/cluster/local/docker-compose.yaml
dweomer: IMO these are too generic for this purpose.

e2e/cluster/local/docker-compose.yaml
dweomer: `--disable` doesn't yet seem to be available in a release

e2e/cluster/local/images/k3s/Dockerfile
galal-hussein: I see, thanks for clarification

e2e/cluster/local/docker-compose.yaml
galal-hussein: got it

e2e/cluster/local/images/k3s/Dockerfile
dweomer: The other reason I need to re-create a k3s image is because I want the tests to run locally without the need to pull anything which means I need the images tarballs in the k3s image. Doing that via docker-compose within Dapper is problematic because the docker daemon it is using does not have the same filesystem root that the host does which means bind mounts are hopeless.

PR #29, opened 2020-02-14
munge plan.status.latestVersion

### LINKED ISSUE #28: Handle k3s semver (v1.17.0+k3s.1)
**Version**

v0.2.0 
**Platform/Architecture**

all
**Describe the bug**

When I try to upgrade to `v1.17.0+k3s.1` the upgrade fails.

**To Reproduce**

Create a plan such as this
```
 apiVersion: upgrade.cattle.io/v1
  kind: Plan
  metadata:
    annotations:
    generation: 3
    name: k3s-up
    namespace: system-upgrade
  spec:
    concurrency: 1
    drain:
      force: true
    nodeSelector:
      matchExpressions:
      - key: k3s-upgrade
        operator: Exists
    serviceAccountName: system-upgrade
    upgrade:
      image: rancher/k3s-upgrade
    version: v1.17.0+k3s.1
```

**Expected behavior**

Upgrade succeeds

**Actual behavior**

Image pulls fails since `rancher/k3s-upgrade:v1.17.0+k3s.1` is not a valid image

dweomer:
This fixes the use case where a channel resolves to (or a version value is) a legitimate SemVer with metadata by replacing all `+` with `-`.
i.e. in the case of k3s, git tag v1.17.2+k3s1 as the version becomes the docker tag v1.17.2-k3s1

Fixes #28


dweomer:
... pleasantly surprised that the added e2e/integration tests to assert the fix worked :tada:

Issue #35, opened 2020-02-21
Nil pointer exception when there is no upgrade field


rmweir:
**Version**

0.2.0

**Platform/Architecture**

linux amd64
**Describe the bug**

bug

**To Reproduce**

apply a plan without an upgrade field

**Expected behavior**

a message indicating it is a no op, or that upgrade is blank

**Actual behavior**

nil pointer exception

dweomer:
Yeah, this definitely needs to be validated. A plan missing an `upgrade` container isn't super useful.

dweomer:
The issue here is that a Plan with `.spec.upgrade` == `nil` is malformed and should be rejected by the api-server. This will require and updated schema. It would be nice if there was a struct annotation that we could use (possibly in the norman project?) to indicate a required field.

dweomer:
As per @ibuildthecloud, a `wrangler:"required"` tag ought to do it.

ShylajaDevadiga:
Verified missing upgrade image, errors out with 
```Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
```

PR #37, opened 2020-02-21
Check if upgrade is nil

rmweir:
Now checks if upgrade is nil. Prior, plans without upgrade would
cause controller to crash.

Issue: https://github.com/rancher/system-upgrade-controller/issues/35


dweomer:
@rmweir I would prefer that the system rejects such an invalid plan rather than running a truncated/useless job. Because the CRD is generated at runtime via norman's openapi integration it this should be as simple as adding the correct annotation on the `Upgrade` member of the `PlanSpec` struct.

PR #47, opened 2020-03-09
reconcile failed jobs with delete

dweomer:
Implements job retry on failure by adapting the wrangler default job reconciler to delete an existing job if it has failed.

Further mitigates rancher/rancher#25882 (by ensuring that upgrades are kept on the treadmill)


dweomer:
@cjellick I don't think I like this one. If we want/need outstanding upgrade jobs to apply and re-apply on nodes until they complete successfully then that should be something that we dial in via controller configuration (or possibly even per-plan if we need that granularity).

Issue #24, opened 2020-02-06
automatic cleanup of upgrade jobs


dweomer:
**Is your feature request related to a problem? Please describe.**
This controller purposely orphans upgrade jobs after they have been scheduled to make reasoning about system state less of a shot-in-the-dark but this can result in jobs and their pods, long since complete, hanging around and polluting `kubectl` output. **We should consider cleaning up stale jobs every so often.**

**Describe the solution you'd like**
Upgrade jobs of a certain age, regardless of the status of plans which were applied, should be cleaned up and removed from the system (this includes their pods). This would probably be most easily accomplished by integrating with the [TTL Controller for Finished Resources](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/)

**Describe alternatives you've considered**
Mark and sweep effected by this controller itself.

**Additional context**
See the `kubectl` output in #23. Pods that incur a reboot as part of the upgrade job will in many cases end up in the `Unknown` state along with a second run of the pod. This may be confusing and shouldn't persist eternally.

tobru:
Good idea! Using the TTL controller could make sense as this doesn't involve any other component which would need to be installed. It certainly would have to be configurable, probably in the Plan spec...

cjellick:
I opened a corresponding issue in rancher for our QA to track: https://github.com/rancher/rancher/issues/25929

dweomer:
Addressed via #48 and #49

PR #48, opened 2020-03-09
implement ttlSecondsAfterFinished for jobs

dweomer:
- bump default backofflimit to the default for the job controller, aka 6

Addresses #24 by implementing cleanup of jobs based on configured .spec.ttlSecondsAfterFinished on jobs.

This should work pretty almost identically to the by-default, gated TTL Controller for Finished Resources.
See https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/


pkg/upgrade/job/job.go
```diff
@@ -16,11 +17,12 @@ import (
...
 const (
-	defaultBackoffLimit          = int32(2)
-	defaultActiveDeadlineSeconds = int64(600)
-	defaultPrivileged            = true
-	defaultKubectlImage          = "rancher/kubectl:1.17.0"
-	defaultImagePullPolicy       = corev1.PullIfNotPresent
+	defaultBackoffLimit            = int32(2)
+	defaultActiveDeadlineSeconds   = int64(600)
+	defaultPrivileged              = true
+	defaultKubectlImage            = "rancher/kubectl:1.17.0"
```
daxmc99: ENV var here also? Might need to drop this down to upgrade a really old cluster

dweomer: that's the only one i have built and the client that system-upgrade-controller matches this so it's kine of moot. if the user needs a lower support version they are on their own (or if they ask nicely, maybe)

pkg/upgrade/handle_batch.go
```diff
@@ -41,10 +53,34 @@ func (ctl *Controller) handleJobs(ctx context.Context) error {
 						}
 					}
 				}
+				if upgradejob.ConditionComplete.IsTrue(obj) {
+					return obj, enqueueOrDelete(jobs, obj, upgradejob.ConditionComplete)
+				}
+				if upgradejob.ConditionFailed.IsTrue(obj) {
+					return obj, enqueueOrDelete(jobs, obj, upgradejob.ConditionFailed)
+				}
```
dweomer: I did purposely leave these under the "if the job has a plan name label" flow control because right now they are the only kind of job supported by this controller but that may not always be the case.

pkg/upgrade/job/job.go
dweomer: Also, this is the default that gets used as a fallback. The provided manifest and hence the rancher/system-charts entry for SUC use a value from a configmap that overrides all of these defaults.

pkg/condition/condition.go
```diff
@@ -0,0 +1,267 @@
...
+	setTS(obj, string(c), ts)
+}
+
+func (c Cond) GetLastUpdated(obj interface{}) string {
+	return getTS(obj, string(c))
+}
+
+func (c Cond) GetLastTransitionTime(obj interface{}) time.Time {
+	return getLastTransitionTime(obj, string(c))
+}
```
dweomer: This entire file is a copy from rancher/wrangler with the addition of `GetLastTransitionTime` because there seems to be no lib out there for interacting with conditions (other than norman or wrangler).

pkg/condition/condition.go
```diff
@@ -0,0 +1,267 @@
+package condition
+
+// adapted from rancher/wrangler
```
daxmc99: Are there a large number of diffs vs wrangler?

pkg/upgrade/handle_batch.go
```diff
@@ -2,24 +2,36 @@ package upgrade
...
-	ctl.batchFactory.Batch().V1().Job().OnChange(ctx, ctl.Name, func(key string, obj *batchv1.Job) (*batchv1.Job, error) {
+	jobs.OnChange(ctx, ctl.Name, func(key string, obj *batchv1.Job) (*batchv1.Job, error) {
 		if obj == nil {
 			return obj, nil
 		}
-		if obj.Labels != nil {
+		jobSelector := labels.SelectorFromSet(labels.Set{
+			upgradeapi.LabelController: ctl.Name,
+		})
+		// avoid commandeering jobs that don't belong to us
```
daxmc99: nit
```suggestion
		// avoid commandeering jobs that do not contain the label for upgrade controller
```

pkg/condition/condition.go
dweomer: Only `GetLastTransitionTime` and `getLastTransitionTime` are new/different.

Issue #59, opened 2020-04-14
Do not force upgrade container tag to be status.latestVersion


masterzen:
This is somehow both a bug report and a feature request.

**Version**

v0.4.0

**Platform/Architecture**

Fedora CoreOS

**Describe the bug**

We were planning to use the System Upgrade Controller to upgrade a fleet of Fedora CoreOS VM. The way the System Upgrade Controller is working fits our model: we could update the plan `version` field with the latest released Fedora CoreOS version (manually or automatically), and the upgrade container would just be an `alpine` container that `chroot` in the image and run the necessary commands to upgrade the OS and reboot to the intended `version`.

Except that it doesn't work because the System Upgrade Controller substitutes the `upgrade` image tag (provided in the plan) by the `status.latestVersion` (ie the plan `version`).
I understand that's how rancher is updated, but that breaks a whole lot of workflows where the upgrade container image is not that important.

I understand that I can workaround this and publish locally an alpine image tagged with the current version (possibly in the `prepare` container, since it's not yet subject to tag overwriting see #41 ), but that'd be awesome, IMHO to not overwrite a specified image tag in both the prepare or upgrade container:

 IMHO, the behavior should be:
* if the `upgrade` or `prepare` container image has the form `image:tag`, then use this explicitely
* if the `upgrade` or `prepare` container image has the form `image` (without a tag) then use the `latestVersion` as the image tag.

This allows both workflows to coexist.

**To Reproduce**

Use the following plan:

```yaml
apiVersion: upgrade.cattle.io/v1
kind: Plan

metadata:
  name: fedora-coreos-asmodee-production
  namespace: system-upgrade
  labels:
    addonmanager.kubernetes.io/mode: Reconcile

spec:
  concurrency: 1
  version: "31.20200323.3.2"
  nodeSelector:
    matchExpressions:
      - {key: role, operator: In, values: ["whatever","dont-care"]}

  serviceAccountName: system-upgrade

  secrets:
    - name: fedora-coreos
      path: /host/run/system-upgrade/fedora-coreos
  upgrade:
    image: alpine:3.11
    command: ["chroot", "/host"]
    args: ["/usr/bin/bash", "/run/system-upgrade/fedora-coreos/upgrade.sh"]
```

**Expected behavior**
It should run the `/run/system-upgrade/fedora-coreos/upgrade.sh` on every node after `chroot`ing to the host.

**Actual behavior**

It fails scheduling the upgrade pod on the nodes with:
```
[...]

Issue #41, opened 2020-03-02
optionally use latest version for prepare container image


dweomer:
**Is your feature request related to a problem? Please describe.**
The `prepare` init container image is static and does not mutate as with the `upgrade` container.  With use-cases such as [k3s-upgrade](https://github.com/rancher/k3s-upgrade) that use essentially the same image for both `prepare` and `upgrade` it would be nice to have a way to let the controller know to use the same tag for `prepare` as is used for `upgrade.

**Describe the solution you'd like**
A way to specify on a **Plan** that the exact same image:tag should be used for the `prepare` container as is used for the `upgrade` container.

**Describe alternatives you've considered**
For **Plans** that have their `version` set explicitly they should also update the `prepare` container to use the same tag as will be used by upgrade.

**Additional context**
n/a

PR #60, opened 2020-04-15
(#59) Do not overwrite container image tag if provided

masterzen:
The current behavior is to force the upgrade container image tag
to be overwritten by status.lastVersion.
While this works for some workflows, it prevents workflows where the
upgrade is done in a non versioned image (or that always requires
latest for instance), because it becomes impossible to mention a
given specific image tag (and especially when using channel instead
of a specific Plan version).

This change implements the following behavior:
* if the image is not specified with a tag, the status.lastVersion
is used, as before.
* if the image specifies a tag, then this tag is used

### DIFF
--- a/pkg/upgrade/container/container.go
+++ b/pkg/upgrade/container/container.go
@@ -46,7 +46,7 @@ func WithImageTag(tag string) Option {
 	return func(container *corev1.Container) {
 		image := container.Image
 		if p := strings.Split(image, `:`); len(p) > 1 {
-			image = strings.Join(p[0:len(p)-1], `:`)
+			return
 		}
 		container.Image = image + `:` + tag
 	}
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ spec:
     force: true
 
   upgrade:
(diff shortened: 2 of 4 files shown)


pkg/upgrade/container/container.go
```diff
@@ -46,7 +46,7 @@ func WithImageTag(tag string) Option {
 	return func(container *corev1.Container) {
 		image := container.Image
 		if p := strings.Split(image, `:`); len(p) > 1 {
-			image = strings.Join(p[0:len(p)-1], `:`)
+			return
```
dweomer, line 49: This should cause some of the test runs for `WithImageTag` to fail. Will merge anyway and fix later.

dweomer:
@masterzen as per my comment above, I realized that my implementation and tests for `WithImageTag` was buggy (because your change should have resulted in failures for the case where a registry was specified with a port). After fixing my implementation and tests I went ahead and implemented what you requested in #59 because it also paired thematically with a very small change to accommodate #41.

Please re-open if you feel that I have missed something in #61

Issue #78, opened 2020-06-02
Pods with a pod disruption budget are not evicted


jcambronero:
**Version**
rancher/system-upgrade-controller:v0.5.0

**Describe the bug**
System upgrade jobs running in hosts that have any pods with a pod disruption budget are stuck forever initializing.

**To Reproduce**
Run any workload that has a pod disruption budget and try to apply an upgrade plan. The job will be stuck on initializing forever on the node where the pod with the pod disruption budget is running.

**Expected behavior**
There should be a way to drain pods with pod disruption budget 

**Actual behavior**
I added the following drain options to my plan:
```
  drain:
    deleteLocalData: true
    ignoreDaemonSets: true
    force: true
```

However, nodes that have any pods with a pod disruption budget are stuck forever initializing. The drain init container is stuck with the following error (that repeats every 5 seconds):

```
evicting pod "<pod-name>"
error when evicting pod "<pod-name>" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```

If I manually remove those pods (using --force --grace-period=0), then the drain container is able to finish and the plan is applied successfully.

dweomer:
Our drains are currently implemented by creating a container that runs on the node in question invoking `kubectl drain`. As of v1.18 it looks like `kubectl drain` has picked up a new flag, `--disable eviction`, for dealing with  scenarios like this. When generating the drain container we will want to perform some minimal check that the kubectl image we are using is v1.18 or newer before including use of this flag. The controller should log a warning when `.spec.drain.disableEvicition` is true but the kubectl image does not support the flag.

A possible alternative is to make use of the `--dry-run=client` flag combined with `--disable-eviction` to see if it is supported.

ShylajaDevadiga:
Using SUC v0.6.1, verified the two scenarios.
1. With minAvailable set to 2 in PDB, on a cluster with 2  schedulable nodes  and concurrency set to 2 and with disableEviction set to true, pods got Evicted and upgrade was successful.
2. With minAvailable set to 2 in PDB, on a cluster with 2 schedulable nodes  and concurrency set to 2 and without disableEviction, jobs were stuck in initializing state and node stuck in Ready,SchedulingDisabled state.

PR #79, opened 2020-06-08
DisableEviction option added

MonzElmasry:
Fixes https://github.com/rancher/system-upgrade-controller/issues/78


README.md
```diff
@@ -129,6 +129,9 @@ spec:
     # deleteLocalData: true  # default
     # ignoreDaemonSets: true # default
     force: true
+    #disableEviction: false #default 
+    #disableEviction flag option is only available in kubectl v1.18 or later, to force drain pods with pod disruption budget.
```
dweomer, line 133: @MonzElmasry as per this comment we should also update the default in the manifest and upgrade/job.go  
- https://github.com/rancher/system-upgrade-controller/blob/v0.5.0/manifests/system-upgrade-controller.yaml#L36
- https://github.com/rancher/system-upgrade-controller/blob/v0.5.0/pkg/upgrade/job/job.go#L23

PR #83, opened 2020-06-10
fix jobs exceeding plan concurrency

### LINKED ISSUE #77: can trigger number of apply jobs greater than plan concurrency
**Version**

v0.3.0+

**Platform/Architecture**

All.

**Describe the bug**

rancher/system-upgrade-controller#38 was only partially mitigated. Refactor [`plan.SelectConcurrentNodeNames()`](https://github.com/rancher/system-upgrade-controller/blob/v0.5.0/pkg/upgrade/plan/plan.go#L123-L178) to add one node at a time to the `${plan}.status.applying`

**To Reproduce**

It depends on internal timing but the most reliable way to reproduce is setup a **Plan** that is ready to apply but prevented by lack of a label then apply the gating label to multiple nodes in a single invocation of `kubectl label`.

**Expected behavior**

Number of running apply jobs for a plan in a cluster does not exceed `${plan}.spec.concurrency`

**Actual behavior**

The number of running apply jobs for a plan in a cluster will typically exceed plan `${plan}.spec.concurrency` roughly 30% of the time.

**Additional context**

See rancher/system-upgrade-controller#39

MonzElmasry:
Fixes #77

### DIFF
--- a/pkg/upgrade/handle_batch.go
+++ b/pkg/upgrade/handle_batch.go
@@ -54,6 +54,17 @@ func (ctl *Controller) handleJobs(ctx context.Context) error {
 		case err != nil:
 			return obj, err
 		}
+		if func(jobNode string, applyingNodes []string) bool {
+			for _, node := range applyingNodes {
+				if jobNode == node {
+					return false
+				}
+			}
+			return true
+		}(obj.Labels[upgradeapi.LabelNode], plan.Status.Applying) {
+			return obj, deleteJob(jobs, obj, metav1.DeletePropagationBackground)
+		}
+
 		// if this job was applying a different version then just delete it
 		// this has the side-effect of only ever retaining one job per node during the TTL window
 		if planVersion != plan.Status.LatestVersion {
--- a/pkg/upgrade/handle_batch.go
+++ b/pkg/upgrade/handle_batch.go
@@ -95,6 +95,7 @@ func (ctl *Controller) handleJobs(ctx context.Context) error {
 			}
 			return obj, enqueueOrDelete(jobs, obj, upgradejob.ConditionComplete)
 		}
+		// if the job is hasn't failed or completed but the job Node is not on the applying list, consider it running out-of-turn and delete it
 		if i := sort.SearchStrings(plan.Status.Applying, nodeName); i == len(plan.Status.Applying) ||
 			(i < len(plan.Status.Applying) && plan.Status.Applying[i] != nodeName) {
 			return obj, deleteJob(jobs, obj, metav1.DeletePropagationBackground)
--- a/pkg/upgrade/handle_batch.go
+++ b/pkg/upgrade/handle_batch.go
@@ -3,6 +3,7 @@ package upgrade
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/upgrade/plan/plan.go
```diff
@@ -184,3 +186,12 @@ func sha256sum(s ...string) string {
 	}
 	return fmt.Sprintf("%x", h.Sum(nil))
 }
+
+func AppendIfMissing(slice []string, element string) []string {
```
briandowns, line 190: Is this function used outside of this package? If not, can you make it unexported?

pkg/upgrade/handle_batch.go
```diff
@@ -54,6 +54,17 @@ func (ctl *Controller) handleJobs(ctx context.Context) error {
 		case err != nil:
 			return obj, err
 		}
+		if func(jobNode string, applyingNodes []string) bool {
```
briandowns, line 57: Can this entire block be simplified?

pkg/upgrade/plan/plan.go
```diff
@@ -168,11 +168,13 @@ func SelectConcurrentNodeNames(plan *upgradeapiv1.Plan, nodeCache corectlv1.Node
 			jsum := sha256sum(string(candidateNodes[j].UID), string(plan.UID), plan.Status.LatestHash)
 			return isum < jsum
 		})
+		for _, node := range applying {
+			selected = AppendIfMissing(selected, node)
+		}
```
dweomer, line 173: I do not see why this block is necessary given the following (from above):
```
		applyingNodes, err := nodeCache.List(nodeSelector.Add(*requirementApplying))
		if err != nil {
			return nil, err
		}
		for _, node := range applyingNodes {
			selected = append(selected, node.Name)
		}
```

Additionally, I believe that this will prevent an "applying" node from ever dropping off of the "selected" slice returned by this func.

pkg/upgrade/handle_batch.go
```diff
@@ -54,6 +54,17 @@ func (ctl *Controller) handleJobs(ctx context.Context) error {
...
+			for _, node := range applyingNodes {
+				if jobNode == node {
+					return false
+				}
+			}
+			return true
+		}(obj.Labels[upgradeapi.LabelNode], plan.Status.Applying) {
+			return obj, deleteJob(jobs, obj, metav1.DeletePropagationBackground)
```
dweomer, line 65: This will prematurely delete a job that has completed successfully (because the node for which it ran will drop off of the "applying" list).

I do not like the idea of deleting a job that that is neither failed nor completed but otherwise meets all of the plan's criteria. **But if we must prune here, because the node that the job is targeting IS NOT on the plan's `.status.applying` list, then it should be the last conditional action of this function.**

Additionally, given that there are no limits placed on the length of `.status.applying` we should eschew looping through all elements looking for a match and instead leverage the fact that the `.status.applying` is sorted which means we can use https://golang.org/pkg/sort/#SearchStrings or https://golang.org/pkg/sort/#Search to perform binary search.

pkg/upgrade/handle_upgrade.go
```diff
@@ -59,7 +59,7 @@ func (ctl *Controller) handlePlans(ctx context.Context) error {
 	)
 
 	// process plan events by creating jobs to apply the plan
-	upgradectlv1.RegisterPlanGeneratingHandler(ctx, plans, ctl.apply.WithCacheTypes(jobs, nodes, secrets).WithNoDelete(), "", ctl.Name,
+	upgradectlv1.RegisterPlanGeneratingHandler(ctx, plans, ctl.apply.WithCacheTypes(nodes, secrets).WithGVK(jobs.GroupVersionKind()).WithDynamicLookup().WithNoDelete(), "", ctl.Name,
```
dweomer, line 62: So this is more correct but it didn't get us anything, is that right @MonzElmasry?

MonzElmasry: yes!

pkg/upgrade/handle_batch.go
```diff
@@ -94,6 +95,10 @@ func (ctl *Controller) handleJobs(ctx context.Context) error {
 			}
 			return obj, enqueueOrDelete(jobs, obj, upgradejob.ConditionComplete)
 		}
```
dweomer, line 97: Would like a comment here, as with the other blocks, explaining what we are deciding to do and why, e.g.:
> // if the job is hasn't failed or completed but is not on the applying list, consider it running out-of-turn and delete it

Issue #88, opened 2020-06-17
drain options: add skip-wait-for-delete-timeout


MonzElmasry:
add skip-wait-for-delete-timeout option to fix unsuccessful deletion of pods because of waitForDelete after the node is labeled for drain.

dweomer:
Fixed via #89

MonzElmasry:
tested via #98

PR #89, opened 2020-06-17
add SkipWaitForDeleteTimeout

MonzElmasry:
#88


pkg/upgrade/job/job.go
```diff
@@ -233,6 +233,9 @@ func New(plan *upgradeapiv1.Plan, nodeName, controllerName string) *batchv1.Job
 			//only available in kubectl version 1.18 or later
 			args = append(args, "--disable-eviction=true")
 		}
+		if drain.SkipWaitForDeleteTimeout > 0 {
```
dweomer, line 236: This should carry the same comment as for `DisableEviction` above, i.e. v1.18 or greater

MonzElmasry: okay, should it be include that in the README as well?

Issue #96, opened 2020-06-25
prevent job creation for unresolved plans


MonzElmasry:
**Version**
v0.5.0 , v0.6.1

**Describe the bug**
Plan generating handler doesn't check if the plan is resolved or failed, which results in generation of invalid job names.

**Expected behavior**
No jobs should be created.

**Actual behavior**
Plan generating handler tries to create jobs with invalid names, so they fail to be created

**Additional context** 
https://github.com/rancher/k3s/issues/1954
https://github.com/rancher/system-upgrade-controller/issues/94

PR #97, opened 2020-06-25
prevent job creation for unresolved plans

MonzElmasry:
- Fix : #96 
- add validation to http response

### DIFF
--- a/e2e/framework/channel.go
+++ b/e2e/framework/channel.go
@@ -7,7 +7,7 @@ import (
 	"os"
 )
 
-func ChannelServer(location string, response int) *httptest.Server {
+func ChannelServer(location string, statusCode int) *httptest.Server {
 	hostname, err := os.Hostname()
 	if err != nil {
 		Failf("cannot read hostname: %v", err)
@@ -15,7 +15,7 @@ func ChannelServer(location string, response int) *httptest.Server {
 	server := &httptest.Server{
 		Config: &http.Server{Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			w.Header().Set("Location", location)
-			w.WriteHeader(response)
+			w.WriteHeader(statusCode)
 		})},
 	}
 	server.Listener, err = net.Listen("tcp", net.JoinHostPort(hostname, "0"))
--- a/e2e/framework/channel.go
+++ b/e2e/framework/channel.go
@@ -7,7 +7,7 @@ import (
 	"os"
 )
 
-func ChannelServer(location string) *httptest.Server {
+func ChannelServer(location string, response int) *httptest.Server {
 	hostname, err := os.Hostname()
 	if err != nil {
 		Failf("cannot read hostname: %v", err)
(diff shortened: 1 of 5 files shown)


pkg/upgrade/plan/plan.go
```diff
@@ -117,7 +117,10 @@ func ResolveChannel(ctx context.Context, url, latestVersion, clusterID string) (
 		}
 		return filepath.Base(redirect.Path), nil
 	}
-	return filepath.Base(url), nil
+	if response.StatusCode < 300 {
```
galal-hussein, line 120: I am not sure if we should select any response code less than 300, this is too general

MonzElmasry: i think it's okay if the response is in 200 range except for 302 which is redirected.  
as per our conversation @dweomer

dweomer: Maybe something like `if (response.StatusCode / 200) == 1 {/* return the base of the passed url */}`m what do you think @galal-hussein ?

pkg/upgrade/handle_upgrade.go
```diff
@@ -65,15 +65,17 @@ func (ctl *Controller) handlePlans(ctx context.Context) error {
...
 			}
 			logrus.Debugf("PLAN GENERATING HANDLER: plan=%s/%s@%s, status=%+v", obj.Namespace, obj.Name, obj.ResourceVersion, status)
-			concurrentNodeNames, err := upgradeplan.SelectConcurrentNodeNames(obj, nodes.Cache())
-			if err != nil {
-				return objects, status, err
-			}
-			logrus.Debugf("concurrentNodeNames = %q", concurrentNodeNames)
-			for _, nodeName := range concurrentNodeNames {
-				objects = append(objects, upgradejob.New(obj, nodeName, ctl.Name))
+			if upgradeapiv1.PlanLatestResolved.IsTrue(obj) {
```
dweomer, line 68: Please instead make the `upgradeapiv1.PlanLatestResolved.IsTrue(obj)` check immediately after the `logrus.Debugf()` line above (which follows the nil check) and reverse it such that if `upgradeapiv1.PlanLatestResolved.IsTrue(obj)` evaluates to `false` then we bail out early and return the same values as the nil check.

pkg/upgrade/plan/plan.go
galal-hussein: yeah sounds good, so basically limiting the responses to 200 and 300

pkg/upgrade/plan/plan.go
```diff
@@ -117,7 +117,10 @@ func ResolveChannel(ctx context.Context, url, latestVersion, clusterID string) (
 		}
 		return filepath.Base(redirect.Path), nil
 	}
-	return filepath.Base(url), nil
+	if (response.StatusCode / 200) == 1 {
+		return filepath.Base(url), nil
+	}
+	return filepath.Base(url), fmt.Errorf("invalid url")
```
dweomer, line 123:
```suggestion
	return "", fmt.Errorf("unexpected response: %s %s", response.Proto, response.Status)
```

e2e/framework/channel.go
```diff
@@ -7,15 +7,15 @@ import (
 	"os"
 )
 
-func ChannelServer(location string) *httptest.Server {
+func ChannelServer(location string, response int) *httptest.Server {
```
dweomer, line 10: Please refactor `response` to be `status` or `statusCode`.

Issue #93, opened 2020-06-23
Fail when update docker package


nickvth:
**Version**
0.0.6

**Platform/Architecture**
Ubuntu

**Describe the bug**
job will fail when updating docker package

**To Reproduce**
apt-get update
apt-get upgrade -y "docker-ce update avaliable"

**Expected behavior**
Update docker package without crashing

**Actual behavior**
Docker daemon crash

**Additional context**
Jun 23 14:41:21 worker-dcg-01 dockerd[1593]: time="2020-06-23T14:41:21.552045334Z" level=info msg="ignoring event" module=libcontainerd namespace=moby topic=/tasks/delete type="*events.TaskDelete"
Jun 23 14:41:21 worker-dcg-01 dockerd[1593]: time="2020-06-23T14:41:21.559750482Z" level=info msg="ignoring event" module=libcontainerd namespace=moby topic=/tasks/delete type="*events.TaskDelete"
Jun 23 14:41:31 worker-dcg-01 dockerd[1593]: time="2020-06-23T14:41:31.258549430Z" level=info msg="Container 1dad5e30dd6e75b2e8c5d9dcd89a064db8ebf6186b6753e95dd5bf9d3a801fed failed to exit within 10 seconds of signal 15 - using the force"
Jun 23 14:41:31 worker-dcg-01 dockerd[1593]: time="2020-06-23T14:41:31.259538935Z" level=info msg="Container cc195360ad0c57786b0e0bd7e1d9a5e5b3e2121da6ce2b4cfcc846dd1e0704a3 failed to exit within 10 seconds of signal 15 - using the force"
Jun 23 14:41:31 worker-dcg-01 dockerd[1593]: time="2020-06-23T14:41:31.267219221Z" level=info msg="Container b188d2811e8ab6ff7375fd932a255571c804c5b29341e6e106d965a57e0b7fd7 failed to exit within 10 seconds of signal 15 - using the force"
Jun 23 14:41:31 worker-dcg-01 dockerd[1593]: time="2020-06-23T14:41:31.351366864Z" level=info msg="ignoring event" module=libcontainerd namespace=moby topic=/tasks/delete type="*events.TaskDelete"
Jun 23 14:41:31 worker-dcg-01 dockerd[1593]: time="2020-06-23T14:41:31.397157353Z" level=info msg="ignoring event" module=libcontainerd namespace=moby topic=/tasks/delete type="*events.TaskDelete"
[...]

Issue #22, opened 2020-02-04
develop examples for general purpose applications


dweomer:
If it's really a general purpose upgrade controller it should work with (and provide examples for):
- [ ] ubuntu (see #19)
- [ ] centos
- [ ] alpine

For each system above provide:
- a reasonably secure, default installation for the controller
- a useable **Plan**
- `kubectl` snippets (or other script assets) for managing application of **Plan**(s)

tobru:
I'll gladly help on this over the next days / weeks.

dweomer:
> I'll gladly help on this over the next days / weeks.

PRs welcome

mitchellmaler:
I wonder if this can be used to orchestrate upgrades for Fedora Core OS as well?

tobru:
> I wonder if this can be used to orchestrate upgrades for Fedora Core OS as well?

Why not? As long as Fedora Core OS can be upgraded on the shell - which it certainly can - then there is a way to do it with SUC :rocket:

tobru:
@dweomer I consider opening a documentation PR for Ubuntu. Where do you want to see this documentation go: In the README? In a separate file f.e. under `docs/example_ubuntu.md`? Do you have any guidelines regarding style (f.e. a boilerplate)?

schmitch:
btw. it would also be great if centos/ubuntu, etc will only be restarted if a certain script renders true or has an exit code of 0, like the centos one: `needs-restarting -r`.

tobru:
@schmitch As the reboot is invoked by the script executed on the node by the Pod started via the Plan, you're absolutely free to implement any logic to decide when to reboot and when not. TL;DR: Yes, it's possible to do what you suggest.

dweomer:
@tobru how does `examples/ubuntu/` directory with a `README.md` and sample manifest(s) sound?

---

(In general, I am fine with folks making their best guess and putting new content where they imagine it belongs and then submitting a PR. For maintainers this can relay useful insight into how users (and other contributors) make use of the project.

dweomer:
@tobru at your convenience, please take a look at the [ubuntu example plan](https://github.com/rancher/system-upgrade-controller/blob/9a801145cca8cf5cd8cf1d3eafb8060c3b9a25b5/examples/ubuntu/bionic.yaml) and let us know what you think.

tobru:
[...]

PR #100, opened 2020-07-03
upgrade plans + documentation update + fix for rancher

robkooper:
This project solved a major headache for me, how to upgrade the underlying OS and cleanly reboot machines.

When experimenting with this I could not easily find in the documentation why the system-upgrade was not running. I added some text to the README to help new users to add the right label to either the master or controlplane nodes.

The cluster I tried this with is managed by rancher and it does not have nodes marked as master, but either as controlplane or etcd. I modified the system-upgrade-controller.yaml to work with these roles as well as the original master role.

Finally my plans to upgrade both bionic and centos 7 machines. This will make sure that the machine reboots and will install docker if needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,14 @@ The most up-to-date manifest is always [manifests/system-upgrade-controller.yaml
 but since release v0.4.0 a manifest specific to the release has been created and uploaded to the release artifacts page.
 See [releases/download/v0.4.0/system-upgrade-controller.yaml](https://github.com/rancher/system-upgrade-controller/releases/download/v0.4.0/system-upgrade-controller.yaml)
 
+Make sure to mark at least one master nodes, or controlplane node with upgrade.cattle.io/controller=system-upgrade-controller:
+
+```shell script
+kubectl label node --selector='node-role.kubernetes.io/master' upgrade.cattle.io/controller=system-upgrade-controller
+# or
+kubectl label node --selector='node-role.kubernetes.io/controlplane' upgrade.cattle.io/controller=system-upgrade-controller
+```
+
 But in the time-honored tradition of `curl ${script} | sudo sh -` here is a nice one-liner:
 
 ```shell script
--- a/manifests/system-upgrade-controller.yaml
+++ b/manifests/system-upgrade-controller.yaml
@@ -58,6 +58,8 @@ spec:
             nodeSelectorTerms:
               - matchExpressions:
                   - {key: "node-role.kubernetes.io/master", operator: In, values: ["true"]}
+              - matchExpressions:
+                  - {key: "node-role.kubernetes.io/controlplane", operator: In, values: ["true"]}
       serviceAccountName: system-upgrade
       tolerations:
         - key: "CriticalAddonsOnly"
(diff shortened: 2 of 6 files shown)


robkooper:
This solves the problem from #93 and adds another example for #22

README.md
```diff
@@ -45,6 +45,14 @@ The most up-to-date manifest is always [manifests/system-upgrade-controller.yaml
...
 See [releases/download/v0.4.0/system-upgrade-controller.yaml](https://github.com/rancher/system-upgrade-controller/releases/download/v0.4.0/system-upgrade-controller.yaml)
 
+Make sure to mark at least one master nodes, or controlplane node with upgrade.cattle.io/controller=system-upgrade-controller:
+
+```shell script
+kubectl label node --selector='node-role.kubernetes.io/master' upgrade.cattle.io/controller=system-upgrade-controller
+# or
+kubectl label node --selector='node-role.kubernetes.io/controlplane' upgrade.cattle.io/controller=system-upgrade-controller
+```
+
```
dweomer, line 55: Under normal circumstances this should not be necessary.

dweomer: Ah, I see now, this goes along with your changes the the controller manifest.

robkooper: When I tried to deploy the manifest for system-upgrade-controller on a system managed by rancher, the pod would never start since there was no node marked with `system-upgrade-controller`

manifests/system-upgrade-controller.yaml
```diff
@@ -58,13 +58,21 @@ spec:
             nodeSelectorTerms:
               - matchExpressions:
                   - {key: "node-role.kubernetes.io/master", operator: In, values: ["true"]}
+              - matchExpressions:
+                  - {key: "node-role.kubernetes.io/controlplane", operator: In, values: ["true"]}
```
zdzichu, line 62: Are you sure about the label? I'm using `k3s` and the label is `node-role.kubernetes.io/control-plane` (notice **-** between word)

robkooper: this is what I see, machines are running CentOS7 with docker and kubernetes installed from rancher.

```
Name:               *****
Roles:              controlplane,etcd
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/os=linux
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=*****
                    kubernetes.io/os=linux
                    node-role.kubernetes.io/controlplane=true
                    node-role.kubernetes.io/etcd=true
Annotations:        node.alpha.kubernetes.io/ttl: 0
                    rke.cattle.io/external-ip: *****
                    rke.cattle.io/internal-ip: *****
                    volumes.kubernetes.io/controller-managed-attach-detach: true
CreationTimestamp:  Mon, 11 Jan 2021 15:05:13 -0600
Taints:             node-role.kubernetes.io/etcd=true:NoExecute
                    node-role.kubernetes.io/controlplane=true:NoSchedule
Unschedulable:      false
```

dweomer: Yeah that was a bug, should have been since corrected.

PR #122, opened 2021-07-26
add control-plane toleration

kinarashah:
We're moving towards using the new control-plane toleration for rke2 addons. Adding control-plane for system-agent so it tolerates both controlplane and control-plane similar to cluster-agent.

https://github.com/rancher/rke2-charts/blob/main-source/packages/rke2-calico/generated-changes/patch/values.yaml.patch#L8 

https://github.com/rancher/rancher/blob/master/pkg/systemtemplate/template.go#L183-L188

https://github.com/rancher/rancher/issues/32904


Oats87 (approved):
LGTM

brandond (approved):

dweomer (approved):

PR #123, opened 2021-07-28
Bump protobuf dependency to latest

cmurphy:


cmurphy:
@dweomer or @kinarashah could you review? I can't add reviewers

dweomer (approved):

dweomer:
LGTM (also @cmurphy you have been added with write permissions which should allow for adding reviewers in the future)

PR #124, opened 2021-08-03
Switch to referencing nodes by hostname label

### LINKED ISSUE #119: Generated node affinity uses cloud-provider hostname instead of k8s hostname
**Version**

`v0.6.2`

**Describe the bug**
When deployed in an RKE cluster managed by Rancher Server, which is affected by [this issue with overriding cloud-provider hostnames](https://github.com/rancher/rancher/issues/18070), the SUC incorrectly generates the `nodeAffinity` for the `kubernetes.io/hostname` label, using the cloud-provider hostname even if it has been overridden within k8s. 

For example, in my shop we do not include our local domain `x-y-z` in our node's hostnames, but since our nodes run in an AWS VPC, the result of `$(curl -s 169.254.169.254/latest/meta-data/local-hostname)` does include the `x-y-z.example.com` domain.

Apparently the SUC reads from this cloud-provider source rather than the k8s API itself, so the resulting node affinities do not match any node, resulting in an unschedulable pod.

EDIT: for clarity, here is the (anonymized) contents of the `Addresses` section of `kubectl describe node x.y.z`
```
Addresses:
  InternalIP:   x.y.z
  Hostname:     ip-x-y-z
  InternalDNS:  ip-x-y-z.example.com
```

**To Reproduce**
1. Deploy an RKE-managed cluster from Rancher Server, using nodes in a cloud provider VPC
2. Locally configure the nodes to have a hostname without the fully-qualified domain present in the official cloud-provider hostname
3. Deploy SUC and a Plan that targets some of the nodes

**Expected behavior**
Plan results in Pods that can actually be scheduled onto the targeted nodes

**Actual behavior**
Plan's resulting Pods have `nodeAffinities` that result in the Pod never being scheduled since they include the fully-qualified `x-y-z.example.com` hostname instead of the overridden hostname.

**Additional context**
Associated unresolved Rancher issue: https://github.com/rancher/rancher/issues/18070

UPDATE: noticed [this section](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#built-in-node-labels) in the k8s docs indicating that the `kubernetes.io/hostname` label is not a reliable means for detecting the node name - in some environments it may differ, so SUC cannot rely on the label to determine node affinities for each Pod.

marcusianlevine:
Closes #119 

Not sure if this will have unintended consequences, but since `node.Name` doesn't always match the `kubernetes.io/hostname` label, this change always reference nodes by their hostname label instead of their `Name`

### DIFF
--- a/pkg/upgrade/plan/plan.go
+++ b/pkg/upgrade/plan/plan.go
@@ -147,7 +147,7 @@ func SelectConcurrentNodeNames(plan *upgradeapiv1.Plan, nodeCache corectlv1.Node
 			return nil, err
 		}
 		for _, node := range applyingNodes {
-			selected = append(selected, node.Name)
+			selected = append(selected, node.Labels[corev1.LabelHostname])
 		}
 		requirementNotApplying, err := labels.NewRequirement(corev1.LabelHostname, selection.NotIn, applying)
 		if err != nil {


dweomer:
@marcusianlevine thank you for the contribution but I do believe that there will be more to this fix. I was settling in to work on it today, actually. Will link here with my findings.

dweomer:
@marcusianlevine I believe that this proposed change is superseded by #126. While it change would work, I think, I had qualms about maintainability and "correctness" (you obviously traced through our code to see that it should work but the slight of hand was too much for me to stomach). My change does break the public-facing API but no one should be using it and I will be moving it to an internal path sometime soon-ish. Please take a look and let me know what you think.

PR #126, opened 2021-08-06
adjust job node affinity to use hostname value

### LINKED ISSUE #119: Generated node affinity uses cloud-provider hostname instead of k8s hostname

dweomer:
Fix job node affinity for nodes with names that do not match the value
of their `kubernetes.io/hostname` label. With this change, node names
are still reported in the **Plan**'s `.status.applying` slice but jobs are generated
with correct node-affinity via matching on the value of
`kubernetes.io/hostname`.

Fixes #119

Signed-off-by: Jacob Blain Christen <jacob@rancher.com>

### DIFF
--- a/pkg/upgrade/handle_upgrade.go
+++ b/pkg/upgrade/handle_upgrade.go
@@ -68,15 +68,14 @@ func (ctl *Controller) handlePlans(ctx context.Context) error {
 			if !upgradeapiv1.PlanLatestResolved.IsTrue(obj) {
 				return objects, status, nil
 			}
-			concurrentNodeNames, err := upgradeplan.SelectConcurrentNodeNames(obj, nodes.Cache())
+			concurrentNodes, err := upgradeplan.SelectConcurrentNodes(obj, nodes.Cache())
 			if err != nil {
 				return objects, status, err
 			}
-			logrus.Debugf("concurrentNodeNames = %q", concurrentNodeNames)
-			for _, nodeName := range concurrentNodeNames {
-				objects = append(objects, upgradejob.New(obj, nodeName, ctl.Name))
+			for _, node := range concurrentNodes {
+				objects = append(objects, upgradejob.New(obj, node, ctl.Name))
+				obj.Status.Applying = append(obj.Status.Applying, node.Name)
 			}
-			obj.Status.Applying = concurrentNodeNames
 			return objects, obj.Status, nil
 		},
 		&generic.GeneratingHandlerOptions{
--- a/pkg/upgrade/job/flags_test.go
+++ b/pkg/upgrade/job/flags_test.go
@@ -5,6 +5,9 @@ import (
 	"testing"
 
 	upgradeapiv1 "github.com/rancher/system-upgrade-controller/pkg/apis/upgrade.cattle.io/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


dweomer:
My only lingering concern with this is that I do not think that there are guards in k8s to prevent multiple nodes having the same value for their `kubernetes.io/hostname` labels which could lead to some random-ish behavior with respect to upgrading. It should still work but it might suffer from being seemingly complete until the job gets garbage-collected by the controller and then re-created because another node has the same name, Additionally, a job could get generated meant for one node but because of the ambiguous node-affinity might land, and complete, on a different node. Hmm...

erikwilson (approved):

Oats87:
I'm thinking that `kubernetes.io/hostname` is too vague of a selector, exactly for the reasons you've specified. I know plenty of instances where hostnames aren't actually accurate or something like `localhost.localdomain` etc.

Is there any other label we can use that would be more unique?

dweomer:
> I'm thinking that `kubernetes.io/hostname` is too vague of a selector, exactly for the reasons you've specified. I know plenty of instances where hostnames aren't actually accurate or something like `localhost.localdomain` etc.
> 
> Is there any other label we can use that would be more unique?

I was hoping for a `kubernetes.io/uid` or the like but seemingly, no dice. It looks as if a KEP may be in order and/or changing SUC to label nodes similarly.

PR #127, opened 2021-08-11
Add ability to redeploy job with generation

StrongMonkey:


dweomer:
@StrongMonkey closing in favor of the existing secret mechanism that we discussed.

PR #128, opened 2021-08-13
Example for nodes using SUSE as an underlying OS

Suse-KevinKlinger:
I've added an example how a Plan looks like if you are running SUSE as the nodes OS.
In this case it's SLES15.


dweomer:
Thanks @Suse-KevinKlinger !

PR #129, opened 2021-08-19
Proposal feature: Adding label to node after applying the job

Suse-KevinKlinger:
This is supposed to be an proposal for a new feature.
It will allow you to add a label to a node you update via your Plans.

Idea:
By this you could label nodes with something like a patch level to distinguish between already updated nodes.
Why would one need this? Imagine you want to add nodes to your cluster and keep all nodes on the same patch level.
You could then setup Plans which exclude already patched nodes via the nodeSelector and apply the updates only to the newly joined nodes.


dweomer:
> Idea:
> By this you could label nodes with something like a patch level to distinguish between already updated nodes.
> Why would one need this? Imagine you want to add nodes to your cluster and keep all nodes on the same patch level.
> You could then setup Plans which exclude already patched nodes via the nodeSelector and apply the updates only to the newly joined nodes.

@Suse-KevinKlinger thank you for starting this conversation with your contribution!

I think that SUC already satisfies this example use-case:

My working assumption for the example use-case that you described is that the "patch level" would be codified in the plan's resolved version. So, something like `v1.2.3` :arrow_right: `v1.2.4` or possibly `15sp2` :arrow_right: `15sp3` (for SLES). Such is already digested in constructing the "latest version" hash, which SUC labels a node with on successful application of an upgrade job. When selecting candidate nodes, informed primarily by the plan's node selector, SUC does also exclude upgraded nodes via this label, e.g. (`plan.upgrade.cattle.io/${plan.name} != ${plan.status.latestVersion}`.

Happy to entertain this further if I am missing the importance of the change for you!

PR #130, opened 2021-08-19
add hard and soft affinity and tolerations

rosskirkpat:
fix for the issue found in https://github.com/rancher/rancher/issues/34282

### DIFF
--- a/pkg/upgrade/job/job.go
+++ b/pkg/upgrade/job/job.go
@@ -137,16 +137,40 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
 						NodeAffinity: &corev1.NodeAffinity{
 							RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
 								NodeSelectorTerms: []corev1.NodeSelectorTerm{{
-									MatchExpressions: []corev1.NodeSelectorRequirement{{
-										Key:      corev1.LabelHostname,
-										Operator: corev1.NodeSelectorOpIn,
-										Values: []string{
-											nodeHostname,
+									MatchExpressions: []corev1.NodeSelectorRequirement{
+										{
+											Key:      corev1.LabelHostname,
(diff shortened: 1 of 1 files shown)


nickgerace (approved):

pkg/upgrade/job/job.go
```diff
@@ -137,16 +137,40 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
...
+											},
 										},
-									}},
+										{
+											Key:      corev1.LabelOSStable,
+											Operator: corev1.NodeSelectorOpIn,
+											Values: []string{
+												"linux",
+											},
+										},
```
dweomer, line 154: This would limit upgrade jobs to only be scheduled on linux nodes. Is that intentional? If so, why would we want that? This doesn't match my understanding of this comment by Denise: https://github.com/rancher/rancher/issues/34282#issuecomment-900689190

pkg/upgrade/job/job.go
```diff
@@ -137,16 +137,40 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
...
+										MatchExpressions: []corev1.NodeSelectorRequirement{
+											{
+												Key:      "node-role.kubernetes.io",
+												Operator: corev1.NodeSelectorOpIn,
+												Values: []string{
+													"control-plane",
+												},
+											}},
+									},
+								}},
```
dweomer, line 171: Is the idea here to favor control-plane nodes for upgrade prior to all other nodes? While that would be an valuable enhancement to what SUC can currently do, please be aware that SUC generates one job per node to which a plan's node-selector applies. Jobs are only ever destined to land on a singular node and so this scheduling preference should never come into play (the node for the generated job has already been picked).

Furthermore, this changes seems superfluous in that, as I understand it, rancher/rancher implements its upgrade controller (that leverages SUC) in such a way that relies on separate, inter-locking plans: one for "servers" and one for "workers" (via node-selector).

PR #131, opened 2021-08-22
Mount host /etc/ca-certificates

RobinMcCorkell:
OSes like Arch Linux put certificate files in /etc/ca-certificates, and symlink to those files from /etc/ssl and /etc/pki.

This is in the same spirit as #116.

### DIFF
--- a/manifests/system-upgrade-controller.yaml
+++ b/manifests/system-upgrade-controller.yaml
@@ -95,6 +95,8 @@ spec:
               mountPath: /etc/ssl
             - name: etc-pki
               mountPath: /etc/pki
+            - name: etc-ca-certificates
+              mountPath: /etc/ca-certificates
             - name: tmp
               mountPath: /tmp
       volumes:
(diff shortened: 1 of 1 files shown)


dweomer (approved):

dweomer:
@RobinMcCorkell I had to spin up an Arch box in Vagrant to verify. So this should enable SUC outbound comms on Arch systems by making the `/etc/ssl/cert.pem` symlink dereferencable for the go runtime. Is that right?

I do worry that the default manifest now has side-effects for just about every system ... for the sake of making it Just Work :tm:  on the very same range of systems. I guess there are worse problems to have.

RobinMcCorkell:
Yep, exactly that.

I wonder if mounting SSL certs into the container is the right approach? Typically when I've built container images I make sure to bundle SSL certs into the image itself (if the app needs to make outbound connections that is). In fact, I believe this is the approach taken by [Google's distroless base image](https://github.com/GoogleContainerTools/distroless).

dweomer:
> I wonder if mounting SSL certs into the container is the right approach? Typically when I've built container images I make sure to bundle SSL certs into the image itself (if the app needs to make outbound connections that is). In fact, I believe this is the approach taken by [Google's distroless base image](https://github.com/GoogleContainerTools/distroless).

A worthy suggestion.

comphilip:
Agree with @RobinMcCorkell, mouting SSL certifications into container is not the approach. In [Flatcar Container Linux](https://kinvolk.io/flatcar-container-linux/), files under `/etc/ssl/certs` are all symbol links to `/usr/share/ca-certificates`

PR #137, opened 2021-09-11
Relax affinity rules for controller

### LINKED ISSUE #136: Controller cannot be scheduled on kuberadm-created clusters
**Version**
v0.7.5

**Platform/Architecture**
* Kubernetes 1.21 via kubeadm on Ubuntu 20.04
* SUC v0.7.5 installed via Kustomize, no adjustments otherwise

**Describe the bug**
The SUC has an affinity rule scheduling the controller onto nodes that are labelled with `node-role.kubernetes.io/master: "true"`.
However, for some versions now the "node-role" labels are applied without a value:

```yaml
apiVersion: v1
kind: Node
metadata:
  name: master-0
  labels:
    node-role.kubernetes.io/master: ""
...
```

Hence the controller cannot be scheduled and stays in "pending"

**To Reproduce**
* Install cluster with kubeadm, deploy SUC

**Expected behavior**
Controller is started on a master system

**Actual behavior**
Controller cannot be scheduled

**Additional context**
FWIW, the cluster runs on Hetzner Cloud and is using the hcloud-cloud-controller-manager.

Should I send a PR to fix the issue?

mhutter:
More recent clusters set an empty value for the node-role labels. This
change makes the SUC work out of the box with those clusters while
retaining backwards compatibility.

Fixes #136

### DIFF
--- a/manifests/system-upgrade-controller.yaml
+++ b/manifests/system-upgrade-controller.yaml
@@ -57,7 +57,7 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
               - matchExpressions:
-                  - {key: "node-role.kubernetes.io/master", operator: In, values: ["true"]}
+                  - {key: "node-role.kubernetes.io/master", operator: Exists}
       serviceAccountName: system-upgrade
       tolerations:
         - key: "CriticalAddonsOnly"


Oats87 (commented):
While I can see the desired use case, this would not work if someone was setting the label value to “false” for whatever reason. Not sure if we should try to handle that.

dweomer:
> While I can see the desired use case, this would not work if someone was setting the label value to “false” for whatever reason. Not sure if we should try to handle that.

```
$ kubectl label node/nuc-1 node-role.kubernetes.io/dweomer=false
$ kubectl get node/nuc-1 --show-labels
NAME    STATUS   ROLES                               AGE    VERSION        LABELS
nuc-1   Ready    control-plane,dweomer,etcd,master   246d   v1.20.7+k3s1   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/instance-type=k3s,beta.kubernetes.io/os=linux,k3os.io/mode=local,k3os.io/upgrade=enabled,k3os.io/version=v0.20.7-k3s1r0,k3s.io/hostname=nuc-1,k3s.io/internal-ip=192.168.1.31,kubernetes.io/arch=amd64,kubernetes.io/hostname=nuc-1,kubernetes.io/os=linux,node-role.kubernetes.io/control-plane=true,node-role.kubernetes.io/dweomer=false,node-role.kubernetes.io/etcd=true,node-role.kubernetes.io/master=true,node.kubernetes.io/instance-type=k3s,plan.upgrade.cattle.io/k3os-latest=dae755315450a65b0ad9a43982fdac70e8dac8a3d7935024f1837d1d

dweomer (approved):

PR #155, opened 2021-09-24
Use plan.Spec.Upgrade in hash

thedadams:
Previously, secrets, image, and service account were used when tracking
updates of a plan with hashing. This made it impossible to trigger an
update by updating something like the environment variables.

After this change, anything in plan.Spec.Upgrade will be used in the
hash to track changes.

Issue:
https://github.com/rancher/rancher/issues/32633
https://github.com/rancher/rancher/issues/34908


thedadams:
@dweomer I ran the tests locally and they passed. However, it seems to have failed here for some reason and I don't have permissions to restart them.

dweomer (approved):

Issue #183, opened 2022-03-02
Manifest file in git is not up to date


sylvainOL:
**Version**
0.8.1

**Describe the bug**
the file `https://github.com/rancher/system-upgrade-controller/blob/master/manifests/system-upgrade-controller.yaml` in branch master and on tag v0.8.1 refers to v.0.8.0 version

The file is OK in the release (kubectl apply -f https://github.com/rancher/system-upgrade-controller/releases/download/v0.8.1/system-upgrade-controller.yaml).

in a "gitops" way, it would be good to have also the file in git synchronized

dweomer:
As hinted in https://github.com/rancher/system-upgrade-controller/pull/184#issuecomment-1072663112 this isn't really an issue as this manifest is the source for what is generated at the time of release.

I would not recommend deploying this component via gitops methodologies unless you like broken clusters.

sylvainOL:
It’s a bit too bad that a gitops tool can’t be installed in a gitops way :-/

chasecaleb:
@dweomer I'd like to second this issue. Even if you don't use "gitops methodologies", this still has downsides:

1. It's confusing. Unexpected surprises like this don't inspire confidence when it's the first thing someone new potentially trips over (exhibit a: me).
2. I use Nix to manage dependencies. If the file is in an actual git repo, I can fetch the branch and Nix will record the commit in its flake lock file -- this way I can use Nix to manage updating dependencies such as this manifest by recreating the lock file. However since this manifest unfortunately isn't in a git repo, I have to use the versioned release url and spend manual time coming back to your GitHub project and checking for a release. I have automation to update my flake lock file, I don't want to reinvent the wheel.

I'd like to use this project, but unfortunately this is a dealbreaker for me due to the maintenance pain. I realize I'm a random drive-by user who hasn't contributed anything to this project, so you don't me anything! I'm just sharing my perspective as a potential user.

Perhaps you could use a GitHub action to update the manifest in master automatically on release, or something like that?

chasecaleb:
Now I'm even more confused because the [current master readme](https://github.com/rancher/system-upgrade-controller/tree/8abd0ddf5051ec6bf5989d7b530e4dc17a55f047#deploying) says you can use kustomize:

> kustomize build github.com/rancher/system-upgrade-controller | kubectl apply -f - 

[...]

PR #184, opened 2022-03-02
Align git manifest with release one (#183)

sylvainOL:
Make the manifest file aligned with the one proposed in release
(https://github.com/rancher/system-upgrade-controller/releases/download/v0.8.1/system-upgrade-controller.yaml).

### DIFF
--- a/manifests/system-upgrade-controller.yaml
+++ b/manifests/system-upgrade-controller.yaml
@@ -12,7 +12,7 @@ metadata:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name:  system-upgrade
+  name: system-upgrade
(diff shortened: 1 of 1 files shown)


dweomer:
Hi @sylvainOL, thank you for your contribution. Scanning this, it seems like a re-format. What problem does this change solve? I've looked at #183 that you have referenced but the underlying issue is that we do not telegraph versions in our manifest which means that a manifest referencing a tagged release will always be at least one commit behind said tag/release.

dweomer:
https://github.com/rancher/system-upgrade-controller/issues/183#issuecomment-1072668808

sylvainOL:
Hi @dweomer, but why don’t you update to the latest version (0.8.1) in master even after the release?

PR #192, opened 2022-03-31
enable fossa scanning

paraglade:
request: https://github.com/rancherlabs/eio/issues/772


dweomer:
@paraglade Looks like if needs to be configured for the PR runners?
Screenshot from 2022-04-21 15-00-42

PR #193, opened 2022-04-12
Migrate to SLE BCI images

thomasferrandiz:
This PR removes the usage of alpine images and replaces them with SLE BCI images.

### DIFF
--- a/e2e/cluster/local/images/k3s/Dockerfile
+++ b/e2e/cluster/local/images/k3s/Dockerfile
@@ -1,12 +1,11 @@
 # Install the all-in-one binary so we can copy our run-time images into the image
 # which helps avoid pulling them when running e2e tests.
-ARG ALPINE="library/alpine:3.14"
-FROM ${ALPINE} AS k3s
+ARG SLES="registry.suse.com/suse/sle15:15.3"
+FROM ${SLES} AS k3s
 ARG ARCH
 ARG K3S_VERSION="v1.21.9+k3s1"
 RUN set -x \
- && apk --no-cache add \
-    bash \
+ && zypper -n in \
     ca-certificates \
     curl \
  && if [ "${ARCH?required}" != "amd64" ]; then \
--- a/.drone.yml
+++ b/.drone.yml
@@ -252,71 +252,6 @@ volumes:
 
 ---
 kind: pipeline
-name: arm
-
-platform:
-  os: linux
-  arch: arm
-
-steps:
-- name: build
-  image: rancher/dapper:v0.5.0
-  commands:
-  - dapper ci
-  volumes:
-  - name: docker
-    path: /var/run/docker.sock
-
-- name: upload-artifacts
-  image: plugins/github-release
-  settings:
-    api_key:
-      from_secret: github_token
-    prerelease: true
-    checksum:
-    - sha256
-    checksum_file: CHECKSUMsum-arm.txt
-    checksum_flatten: true
-    files:
-    - "dist/artifacts/*"
-  when:
-    instance:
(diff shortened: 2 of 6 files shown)


pgonin:
poke @macedogm

pgonin:
build fail because of `no matching manifest for linux/arm/v7 in the manifest list entries`
I don't think we need to keep building for arm v7

macedogm:
@pgonin I spoke with Caleb about `armv7` and we don't need to build for this arch, so it's safe to remove it from the build pipeline. I had do to it also for some of the PRs that I opened. We need to build and ship only `amd64`, `s390x` and `arm64`.

macedogm (changes requested):
We just need to remove the build step for `armv7` from Drone, then LGTM.

dweomer:
> @pgonin I spoke with Caleb about `armv7` and we don't need to build for this arch, so it's safe to remove it from the build pipeline. I had do to it also for some of the PRs that I opened. We need to build and ship only `amd64`, `s390x` and `arm64`.

This is incorrect: SUC is leveraged to upgrade k3s which still ships for arm v7.

e2e/cluster/local/images/k3s/Dockerfile
```diff
@@ -1,12 +1,11 @@
 # Install the all-in-one binary so we can copy our run-time images into the image
 # which helps avoid pulling them when running e2e tests.
-ARG ALPINE="library/alpine:3.14"
-FROM ${ALPINE} AS k3s
+ARG SLES="registry.suse.com/suse/sle15:15.3"
```
dweomer, line 3: Why SLES here instead of BCI? Unpacked on my system, the SLES image is 7 times larger than Alpine.

dweomer:
This is also a go version bump. In trying to see if the go.mod/go.sum would change with a `go mod tidy` I ran into:
```
$ go mod tidy
go: downloading github.com/docker/distribution v2.8.0+incompatible
verifying github.com/docker/distribution@v2.8.0+incompatible: checksum mismatch
        downloaded: h1:u9vuu6qqG7nN9a735Noed0ahoUm30iipVRlhgh72N0M=
        go.sum:     h1:l9EaZDICImO1ngI+uTifW+ZYvvz7fKISBAKpg+MbWbY=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
```

Fixed via https://github.com/rancher/system-upgrade-controller/pull/196

e2e/cluster/local/images/k3s/Dockerfile
thomasferrandiz: In fact it looks like `bci-base` and `sle15` images are identical.
If you check here, they have the same digest:
https://registry.suse.com/static/suse/sle15sp3/index.html
https://registry.suse.com/static/bci/bci-base/index.html

But I agree it is unfortunate that they are so large.

There are also `bci-minimal` and `bci-micro` images which are small but they are still in Tech Preview.
As I understand, @macedogm is preparing a proposal to use them with sidecars but I don't know if we can use that yet.

macedogm: I'm discussing internally about which image to "standardize". I don't believe that we will follow with the idea of the sidecar, but actually with a mix of `bci-micro` and `bci-base/sle15` (when really needed). Let's see how it'll progress.

cbron: `registry.suse.com/suse/sle15:15.3` is sle-bci, not full SLES.

PR #195, opened 2022-04-21
Add Volume Mounts

luthermonson:
Add's a new Volume struct which we use to add pod/container volumes. When we get them in the SUC we can then loop over and attach to all containers in the pod which will allow to let the plan decide what it needs for volumes. This will be used by the windows wins upgrade plan to mount the wins npipe socket into the container so we can break out and run wins commands.

https://github.com/rancher/rancher/issues/37020


dweomer:
As discussed in Slack:

I think what we need here is a new `VolumeSpec` (or `MountSpec` or even `BindSpec`) sub-type, akin to our custom `SecretSpec`, that references a host location/source (i.e. `HostPath` or `Source`) to mount into the containers, e.g.:
```go
type MountSpec struct {
	Name     string `json:"name,omitempty"`
	Path     string `json:"path,omitempty"`
	HostPath string `json:"path,omitempty"`
}
```

---

I need to think about it more to decide if I am comfortable with adding general purpose volumes to plans. :thinking:

pkg/upgrade/container/container.go
```diff
@@ -110,6 +110,11 @@ func New(name string, spec upgradeapiv1.ContainerSpec, opt ...Option) corev1.Con
 		}}, spec.Env...),
 		EnvFrom: spec.EnvFrom,
 	}
+
+	for _, v := range spec.VolumeMounts {
+		container.VolumeMounts = append(container.VolumeMounts, v)
+	}
+
```
dweomer, line 117: This isn't enough because you can reference volumes by name all day but if they do not exist for the pod ...

dweomer (changes requested):
As discussed in Slack:

I think what we need here is a new `VolumeSpec` (or `MountSpec` or even `BindSpec`) sub-type, akin to our custom `SecretSpec`, that references a host location/source (i.e. `HostPath` or `Source`) to mount into the containers, e.g.:
```go
type MountSpec struct {
	Name     string `json:"name,omitempty"`
	Path     string `json:"path,omitempty"`
	HostPath string `json:"path,omitempty"`
}
```

You could use such in a plan to give you enough information to specify both the volumes for the pod and the references to such for individual containers (assuming you need to reference them: maybe we just mount any volumes specified for all containers?).

dweomer:
> As discussed in Slack:
> 
> I think what we need here is a new `VolumeSpec` (or `MountSpec` or even `BindSpec`) sub-type, akin to our custom `SecretSpec`, that references a host location/source (i.e. `HostPath` or `Source`) to mount into the containers, e.g.:
> 
> ```go
> type MountSpec struct {
> 	Name     string `json:"name,omitempty"`
> 	Path     string `json:"path,omitempty"`
> 	HostPath string `json:"path,omitempty"`
> }
> ```
> 
> You could use such in a plan to give you enough information to specify both the volumes for the pod and the references to such for individual containers (assuming you need to reference them: maybe we just mount any volumes specified for all containers?).

I need to think about it more to decide if I am comfortable with adding general purpose volumes to plans. :thinking:

luthermonson:
ya i agree, there is more to this than just volume mounts... le sigh

pkg/apis/upgrade.cattle.io/v1/types.go
```diff
@@ -62,6 +62,13 @@ type ContainerSpec struct {
 	Args    []string               `json:"args,omitempty"`
 	Env     []corev1.EnvVar        `json:"envs,omitempty"`
 	EnvFrom []corev1.EnvFromSource `json:"envFrom,omitempty"`
+	Volumes []Volume               `json:"volumes,omitempty"`
+}
+
+type Volume struct {
```
dweomer, line 68:
```suggestion
type VolumeSpec struct {
```

pkg/apis/upgrade.cattle.io/v1/types.go
```diff
@@ -62,6 +62,13 @@ type ContainerSpec struct {
...
+	Name        string
+	Source      string
+	Destination string
 }
```
dweomer, line 72:
```suggestion
type VolumeSpec struct {
	Name        string `json:"name,omitempty"`
	Source      string `json:"source,omitempty"`
	Destination string `json:"destination,omitempty"`
}
```

dweomer (commented):
Looks good (minus the nit/suggestion hinting at stylistic divergence).

pkg/upgrade/job/job.go
```diff
@@ -212,6 +212,18 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
 		})
 	}
 
+	// add volumes from upgrade plan
+	for _, v := range plan.Spec.Upgrade.Volumes {
+		podTemplate.Spec.Volumes = append(podTemplate.Spec.Volumes, corev1.Volume{
+			Name: v.Name,
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: v.Source, Type: &hostPathDirectory,
```
dweomer, line 221:
```
					Path: v.Source, Type: &hostPathDirectory,
```

:thinking: I thought that we wanted to leave the type off if we could?

PR #231, opened 2023-03-15
Add support for PodSelector in DrainSpec

brandond:
The selected labels are added to the default LabelSelector that excludes the SUC pods from being drained.

Also bumps various go modules and the kubectl image up to latest for Kubernetes 1.21

* SURE-5988
* SURE-5994
* SURE-6032
* https://github.com/rancher/rancher/issues/40003


briandowns (approved):

Oats87 (approved):

pkg/apis/upgrade.cattle.io/v1/types.go
```diff
@@ -73,13 +73,15 @@ type VolumeSpec struct {
...
-	GracePeriod              *int32         `json:"gracePeriod,omitempty"`
-	DeleteLocalData          *bool          `json:"deleteLocalData,omitempty"`
-	IgnoreDaemonSets         *bool          `json:"ignoreDaemonSets,omitempty"`
-	Force                    bool           `json:"force,omitempty"`
-	DisableEviction          bool           `json:"disableEviction,omitempty"`
-	SkipWaitForDeleteTimeout int            `json:"skipWaitForDeleteTimeout,omitempty"`
+	Timeout                  *time.Duration        `json:"timeout,omitempty"`
+	GracePeriod              *int32                `json:"gracePeriod,omitempty"`
+	DeleteLocalData          *bool                 `json:"deleteLocalData,omitempty'"`
+	DeleteEmptydirData       *bool                 `json:"deleteEmptydirData,omitempty"`
```
snasovich, line 79: Looks like `DeleteEmptydirData` is essentially "alias" for `DeleteLocalData`. I understand the `DeleteEmptydirData` is more appropriate name for what the setting is for now and we can't remove/rename field for backward compatibility purposes.

But adding this field adds the potential to cause confusion for SUC user where they can set one of `DeleteEmptydirData` and `DeleteLocalData` to `true` and the other to `false` and expecting something that SUC doesn't actually support.

So my recommendation is to leave things be "as is" and not add new `DeleteEmptydirData` field.

@Oats87 @brandond , WDYT?

rancher-max:
https://github.com/k3s-io/k3s/issues/7099
https://github.com/rancher/rke2/issues/4021

pkg/apis/upgrade.cattle.io/v1/types.go
brandond: The flag has been renamed since Kubernetes 1.20 and the old flag is totally gone from the upstream docs. I'd prefer to start transitioning over to the proper name rather than just stay with something that is going to be confusing for anyone approaching this with no knowledge of what this actually maps to on modern versions of Kubernetes. I do agree that the current behavior of just silently allowing conflicting settings could lead to confusion. Let me take a shot at addressing that.

PR #237, opened 2023-04-21
examples: add upgrade YAML for MicroOS

aplanas:


aplanas:
This PR is ready for review

aplanas:
I tested this, and I found one issue with selinux (bsc#1210822) and k3s-selinux (bsc#1210824). Only the first one is important, but this plan has been tested and drive the installation until the end.

aplanas:
cc @brandond

brandond (approved):

examples/openSUSE/microos.yaml
```diff
@@ -0,0 +1,40 @@
...
+apiVersion: upgrade.cattle.io/v1
+kind: Plan
+metadata:
+  name: microos
+  namespace: system-upgrade
+spec:
+  concurrency: 2
+  nodeSelector:
+    matchExpressions:
+      - {key: node-role.kubernetes.io/worker, operator: In, values: ["true"]}
```
brandond, line 29: Are you sure about this? RKE2 and K3s do not use the 'worker' role so this will never match any nodes.

aplanas: I label it as such, but I used the same that is in https://github.com/rancher/system-upgrade-controller/blob/master/examples/suse/sles.yaml#L26.

Can something like this be better? https://github.com/rancher/system-upgrade-controller/blob/master/examples/ubuntu/bionic.yaml#L31

brandond: I don't know, none of these labels will exist by default - I guess the examples assume people understand that they need to label their nodes to match the LabelSelector in the plan? I don't mean to make you fix all of this; I approved so you can merge if you want .

aplanas: Excellent. I do not have rights to merge it. Can you do that?

Thanks!

PR #244, opened 2023-06-21
Allow Plan to specify Job deadline

connorkuehl:
If a workload is known to be slow in excess of the system-upgrade-controller deployment's default deadline, allow the Plan creator to specify a job deadline with the plan so that the entailing job may run to completion.

### DIFF
--- a/pkg/apis/upgrade.cattle.io/v1/types.go
+++ b/pkg/apis/upgrade.cattle.io/v1/types.go
@@ -33,9 +33,10 @@ type Plan struct {
 
 // PlanSpec represents the user-configurable details of a Plan.
 type PlanSpec struct {
-	Concurrency        int64                 `json:"concurrency,omitempty"`
-	NodeSelector       *metav1.LabelSelector `json:"nodeSelector,omitempty"`
-	ServiceAccountName string                `json:"serviceAccountName,omitempty"`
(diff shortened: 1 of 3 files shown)


connorkuehl:
@brandond Would you be willing to take a look at this?

brandond (changes requested):
This is great, thanks for the changes and matching tests!

I feel like we should allow the administrator to specify an upper bound on the deadline - perhaps add a `ActiveDeadlineSecondsMax` var wired up to `SYSTEM_UPGRADE_JOB_ACTIVE_DEADLINE_SECONDS_MAX`, that (if set) provides an upper bound for the value from the job spec? That gives the cluster administrator an option to cap the user-provided value.

brandond (approved):

connorkuehl:
Thanks for the review! :slightly_smiling_face:

briandowns (approved):

dereknola (approved):

connorkuehl:
I'm still learning about the Rancher ecosystem, so at the risk of asking a dumb question: is there a way for me to track when this change is packaged up as part of a [Rancher release](https://github.com/rancher/rancher/releases)?

brandond:
If there's not an issue to upgrade the version of the SUC used by Rancher's upgrade charts, one will need to be created to track doing so. Standalone users can use the new release whenever they want, of course.

PR #257, opened 2023-08-24
Adding the ability to define a secuirty context and SELinux options

Auston-Ivison-Suse:
We were running into the problems with the upgrade job failing on a host with SELinux enabled. So when one would need to upgrade a host's OS, allowing to set a security context and SELinux Options therein will allow you to upgrade the OS, without disabling SELinux beforehand. 

Within the yaml file of a plan, you will be able to set `plan.Spec.Upgrade.SecurityContext` and it will take effect in the upgrade job. The same concept applies to the SELinux Options within a plan.


pkg/apis/upgrade.cattle.io/v1/types.go
```diff
@@ -60,12 +60,14 @@ type PlanStatus struct {
 
 // ContainerSpec is a simplified container template.
 type ContainerSpec struct {
-	Image   string                 `json:"image,omitempty"`
-	Command []string               `json:"command,omitempty"`
-	Args    []string               `json:"args,omitempty"`
-	Env     []corev1.EnvVar        `json:"envs,omitempty"`
-	EnvFrom []corev1.EnvFromSource `json:"envFrom,omitempty"`
-	Volumes []VolumeSpec           `json:"volumes,omitempty"`
+	Image   string                 				`json:"image,omitempty"`
```
brandond, line 63: please `go fmt` this properly; the indentation is unusual.

brandond (changes requested):
Are you going to plumb these new fields through to anything? See how the SecurityContext for the upgrade pod is generated at https://github.com/rancher/system-upgrade-controller/blob/04a0b9ef5858657f20949cd022e58ad19de029df/pkg/upgrade/job/job.go#L326-L333

You need to modify that bit to actually use the SecurityContext from the Plan, if it is non-nil. You also need to similarly wire up the SELinuxOptions.

brandond:
Also, please fill out the PR body with some background and an example of how and when this might be used.

pkg/upgrade/container/container.go
```diff
@@ -37,6 +37,12 @@ func WithSecurityContext(securityContext *corev1.SecurityContext) Option {
 	}
 }
 
+func WithSELinuxOptions(seLinuxoptions *corev1.SELinuxOptions) Option {
```
brandond, line 40: I'm not sure this is necessary - you're setting `container.SecurityContext.SELinuxOptions`, when `container.SecurityContext` is already set by the `WithSecurityContext` function. Just set this directly on the SecurityContext.

pkg/apis/upgrade.cattle.io/v1/types.go
```diff
@@ -60,12 +60,14 @@ type PlanStatus struct {
...
-	EnvFrom []corev1.EnvFromSource `json:"envFrom,omitempty"`
-	Volumes []VolumeSpec           `json:"volumes,omitempty"`
+	Image           string                  `json:"image,omitempty"`
+	Command         []string                `json:"command,omitempty"`
+	Args            []string                `json:"args,omitempty"`
+	Env             []corev1.EnvVar         `json:"envs,omitempty"`
+	EnvFrom         []corev1.EnvFromSource  `json:"envFrom,omitempty"`
+	Volumes         []VolumeSpec            `json:"volumes,omitempty"`
+	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
+	SELinuxOptions  *corev1.SELinuxOptions  `json:"seLinuxOptions,omitempty"`
```
brandond, line 70: Just realized that SELinuxOptions is actually nested within the SecurityContext struct; we don't need a separate field for it. Is there any specific reason why you wanted to allow setting it separately? I guess I could see the value in allowing use of the default SecurityContext, but just overriding the SELinuxOptions. I could see this being confusing though, as it provides two places to set the SELinuxOptions - both in the SecurityContext's nested field, and in the top-level ContainerSpec field.

pkg/upgrade/job/job.go
```diff
@@ -319,18 +319,31 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
...
+	}
+	if plan.Spec.Upgrade.SecurityContext != nil {
+		securityContext = plan.Spec.Upgrade.SecurityContext
+	}
+
+	// Check if SELinuxOptions from the Plan is non-nil
+	seLinuxOptions := &corev1.SELinuxOptions{}
+	if plan.Spec.Upgrade.SELinuxOptions != nil {
+		seLinuxOptions = plan.Spec.Upgrade.SELinuxOptions
+	}
```
brandond, line 339:
```suggestion
	if plan.Spec.Upgrade.SELinuxOptions != nil {
		securityContext.SELinuxOptions = plan.Spec.Upgrade.SELinuxOptions
	}
```

pkg/upgrade/job/job.go
```diff
@@ -319,18 +319,31 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
...
-			upgradectr.WithSecurityContext(&corev1.SecurityContext{
-				Privileged: &Privileged,
-				Capabilities: &corev1.Capabilities{
-					Add: []corev1.Capability{
-						corev1.Capability("CAP_SYS_BOOT"),
-					},
-				},
-			}),
+			upgradectr.WithSecurityContext(securityContext),
+			upgradectr.WithSELinuxOptions(seLinuxOptions),
```
brandond, line 346:
```suggestion
```

brandond (changes requested):
Couple nits on how this is plumbed through to the `upgrade` container.

Note that the `prepare` container also uses all the fields from `ContainerSpec`; we should probably wire up all the SecurityContext stuff there as well. I'm not sure why it currently doesn't do this, as it's not documented as only supported for the upgrade pod.

pkg/upgrade/job/job.go
```diff
@@ -319,18 +320,28 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
...
+		},
+	}
+	if plan.Spec.Upgrade.SecurityContext != nil {
+		securityContext = plan.Spec.Upgrade.SecurityContext
+	}
+
+	if plan.Spec.Upgrade.SecurityContext.SELinuxOptions != nil {
+		securityContext.SELinuxOptions = plan.Spec.Upgrade.SecurityContext.SELinuxOptions
+	}
+
```
brandond, line 339: This is redundant. The only way to set the SELinuxOptions is via a non-nil plan.Spec.Upgrade.SecurityContext, which we just assigned into securityContext in the previous check.
```suggestion
```

brandond (changes requested):
Currently, the option to use a privileged pod for the upgrade pod is set at the controller level:
https://github.com/rancher/system-upgrade-controller/blob/f4a9a0fbe69b144b5b6f6cf69aab5472498211b0/pkg/upgrade/job/job.go#L73-L82

Since this is a security control enforced by the administrator, I feel like we probably also need to wire up a new option to allow administrators to opt in to allowing users to specify a SecurityContext for their pods. The option should default to false, and if it is false the SecurityContext from the Plan should be ignored - perhaps along with a logged warning?

brandond:
@Auston-Ivison-Suse are you still working on this?

brandond:
You generally want to rebase your branch on top of master, instead of pulling master into your branch. makes for a cleaner commit log.

dweomer (approved):

brandond (approved):

Issue #225, opened 2023-02-22
Some way to prevent concurrent plans on the cluster


davidcassany:
**Is your feature request related to a problem? Please describe.**
The problem we are facing is finding a proper way to prevent two different upgrade plans to be executed concurrently on the same cluster or even on the same node. In Elemental we found nothing prevents multiple upgrade plans to be executed concurrently over the same node. This has potential to corrupt the node OS.

**Describe the solution you'd like**
Having some sort of configuration to prevent simultaneous plans being executed. Once a plan started in a cluster another one can't be scheduled until the current ongoing one has ended.

**Describe alternatives you've considered**
I have been wondering of setting on some workarounds at node level like setting taints in prepare step and then setting some tolerations to prevent having multiple plans accessing the same node simultaneously and after thinking about and digging I believe this is not right, as I need something at cluster level. Preventing concurrent plans on  a node can lead to scenarios where the plans are executed in different order depending on the node. I think there should be some way to prevent that from happening.

**Additional context**
This issue feels like it could be related to #163, but since the use case is different I opened a new issue in case I missed something, I am still learning how system-upgrade-controller works.

PR #260, opened 2023-08-29
225: Add support for exclusive plans

jrodonnell:
Add new field to Plan spec, Exclusive. This field will be useful when the order of execution for a series of Jobs is important, as each exclusive Job will have to complete before the next one is scheduled.

Setting the field to true will change the default behavior of how Job Pods are scheduled: it will create a PodAntiAffinity rule which will prevent more than one `exclusive` labeled SUC Job Pod from being scheduled to a node at a time, and will add this label to the Job Pod. 

If omitted or explicitly set to false, Job Pods will be scheduled according to current default behavior. 

Implements request in #225.

@dweomer

### DIFF
--- a/pkg/apis/upgrade.cattle.io/constants.go
+++ b/pkg/apis/upgrade.cattle.io/constants.go
@@ -15,7 +15,7 @@ const (
 	// LabelController is the name of the upgrade controller.
 	LabelController = GroupName + `/controller`
 
-	// LabelExclusive is the value of the Exclusive flag.
+	// LabelExclusive is set if the plan should not run concurrent with other plans.
 	LabelExclusive = GroupName + `/exclusive`
 
 	// LabelNode is the node being upgraded.
--- a/pkg/apis/upgrade.cattle.io/constants.go
+++ b/pkg/apis/upgrade.cattle.io/constants.go
@@ -15,6 +15,9 @@ const (
 	// LabelController is the name of the upgrade controller.
 	LabelController = GroupName + `/controller`
 
+	// LabelExclusive is the value of the Exclusive flag.
+	LabelExclusive = GroupName + `/exclusive`
+
 	// LabelNode is the node being upgraded.
 	LabelNode = GroupName + `/node`
 
--- a/pkg/apis/upgrade.cattle.io/v1/types.go
+++ b/pkg/apis/upgrade.cattle.io/v1/types.go
@@ -44,6 +44,8 @@ type PlanSpec struct {
 
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
 
+	OnePlanPerNode bool `json:"onePlanPerNode,omitempty"`
+
 	Prepare *ContainerSpec `json:"prepare,omitempty"`
 	Cordon  bool           `json:"cordon,omitempty"`
 	Drain   *DrainSpec     `json:"drain,omitempty"`
--- a/pkg/apis/upgrade.cattle.io/v1/types.go
+++ b/pkg/apis/upgrade.cattle.io/v1/types.go
@@ -44,7 +44,7 @@ type PlanSpec struct {
 
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
 
-	OnePlanPerNode bool `json:"onePlanPerNode,omitempty"`
+	Exclusive bool `json:"exclusive,omitempty"`
 
 	Prepare *ContainerSpec `json:"prepare,omitempty"`
 	Cordon  bool           `json:"cordon,omitempty"`
--- a/pkg/upgrade/job/job.go
+++ b/pkg/upgrade/job/job.go
@@ -161,13 +161,7 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
 						PodAntiAffinity: &corev1.PodAntiAffinity{
 							RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{{
 								LabelSelector: &metav1.LabelSelector{
-									MatchExpressions: []metav1.LabelSelectorRequirement{{
-										Key:      upgradeapi.LabelPlan,
-										Operator: metav1.LabelSelectorOpIn,
-										Values: []string{
-											plan.Name,
-										},
-									}},
+									MatchExpressions: onePlanPerNode(plan, node),
 								},
 								TopologyKey: corev1.LabelHostname,
 							}},
@@ -358,4 +352,31 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
 	}
 
 	return job
+}
+
+func onePlanPerNode(plan *upgradeapiv1.Plan, node *corev1.Node) []metav1.LabelSelectorRequirement {
+	var (
+		oppn     = plan.Spec.OnePlanPerNode
+		selector []metav1.LabelSelectorRequirement
+	)
+
+	if oppn == true {
+		selector = []metav1.LabelSelectorRequirement{{
+			Key:      upgradeapi.LabelNode,
+			Operator: metav1.LabelSelectorOpIn,
+			Values: []string{
+				node.Name,
+			},
+		}}
+	} else {
+		selector = []metav1.LabelSelectorRequirement{{
+			Key:      upgradeapi.LabelPlan,
+			Operator: metav1.LabelSelectorOpIn,
+			Values: []string{
+				plan.Name,
+			},
+		}}
+	}
+
+	return selector
 }
--- a/pkg/upgrade/job/job.go
+++ b/pkg/upgrade/job/job.go
@@ -161,7 +161,13 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
 						PodAntiAffinity: &corev1.PodAntiAffinity{
 							RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{{
 								LabelSelector: &metav1.LabelSelector{
-									MatchExpressions: onePlanPerNode(plan, node),
+									MatchExpressions: []metav1.LabelSelectorRequirement{{
+										Key:      upgradeapi.LabelPlan,
+										Operator: metav1.LabelSelectorOpIn,
+										Values: []string{
+											plan.Name,
+										},
+									}},
 								},
 								TopologyKey: corev1.LabelHostname,
 							}},
@@ -203,6 +209,21 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
 	*job.Spec.Completions = 1
 	if i := sort.SearchStrings(plan.Status.Applying, nodeHostname); i < len(plan.Status.Applying) && plan.Status.Applying[i] == nodeHostname {
 		*job.Spec.Parallelism = 1
+	}
+
+	if plan.Spec.OnePlanPerNode {
+		job.Spec.Template.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution = []corev1.PodAffinityTerm{{
+			LabelSelector: &metav1.LabelSelector{
+				MatchExpressions: []metav1.LabelSelectorRequirement{{
+					Key:      upgradeapi.LabelNode,
+					Operator: metav1.LabelSelectorOpIn,
+					Values: []string{
+						node.Name,
+					},
+				}},
+			},
+			TopologyKey: corev1.LabelHostname,
+		}}
 	}
 
 	podTemplate := &job.Spec.Template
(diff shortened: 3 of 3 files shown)


dweomer (approved):
This looks fine to me but please consider my comments: I do not think the `onePlanPerNode` function is necessary and it obscures the retention of the existing default behavior and that the added bool on the Plan is opt-in.

brandond:
I'm confused as to why this is necessary; isn't the proscribed way to do this via something like the k3s-upgrade "prepare" step, where each subsequent plan waits for the previous plan to complete before moving into the upgrade phase?

> This behavior will be useful when the order of execution for a series of Jobs is important, as the Job will have to complete before the next one is scheduled.

If the desire is to bring this closer to full workflow engine with strict sequencing and inter-job dependency tracking built into the controller, instead of leaning on the images to figure it out for themselves, just slapping a flag on the Plan that prevents it from running alongside any other Plan doesn't seem like it gets us there?

pkg/apis/upgrade.cattle.io/v1/types.go
```diff
@@ -44,6 +44,8 @@ type PlanSpec struct {
 
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
 
+	OnePlanPerNode bool `json:"onePlanPerNode,omitempty"`
```
brandond, line 47: is `Exclusive` perhaps a better name for this option? `OnePlanPerNode` is a bit of a mouthful.

pkg/upgrade/job/job.go
```diff
@@ -161,13 +161,7 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
...
 							RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{{
 								LabelSelector: &metav1.LabelSelector{
-									MatchExpressions: []metav1.LabelSelectorRequirement{{
-										Key:      upgradeapi.LabelPlan,
-										Operator: metav1.LabelSelectorOpIn,
-										Values: []string{
-											plan.Name,
-										},
-									}},
+									MatchExpressions: onePlanPerNode(plan, node),
```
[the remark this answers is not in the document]
brandond: Yeah, I don't like replacing the entire expression here with a call to `onePlanPerNode`, as the function is responsible for generating the LabelSelectorRequirement regardless of whether or not the new option is in use. Can we leave this as-is and either mutate or add to the `[]metav1.LabelSelectorRequirement` later, if the option is set?

brandond (changes requested):
A couple nits on the field name and expression generation logic.

Just to be clear, the intent is for this option to only force exclusivity/serialization among plans with the option set. Only a single "exclusive" or "serialized" plan may run at a time, alongside zero or more "non-exclusive" plans.

jrodonnell:
> Just to be clear, the intent is for this option to only force exclusivity/serialization among plans with the option set. Only a single "exclusive" or "serialized" plan may run at a time, alongside zero or more "non-exclusive" plans.

I believe the intent was to make an exclusive Plan block any other Plan from running, regardless of any other Plan's settings, for a few reasons. I see this exclusivity as an important statement of fact and intent, that running any other Plans while this exclusive one is running is a bad idea because it will be doing things to the node that could break other Jobs in the middle of execution and thus leave the node in an unknown state, and allowing that declaration to be effectively overridden or ignored by other Plans would seem to defeat that purpose. Under those same assumptions, requiring all Plans to have the flag in order to comply with one Plan’s exclusivity requirement would be counterintuitive to me and would also force all Plans into serial execution even if only one of them actually needed to be exclusive (or force the operator into a two-step process: apply exclusive Plans, wait to finish, then apply normal ones).

Whereas if the exclusive Plans just block all other Plans while they are running, then when they are done the rest will just automatically execute according to the current default scheduling. To me that seems like the most intuitive behavior that operators can easily understand and leverage in practice.

pkg/apis/upgrade.cattle.io/v1/types.go
jrodonnell: I agree, I do like `Exclusive`, I think that is concise.

pkg/upgrade/job/job.go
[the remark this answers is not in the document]
jrodonnell: I agree that makes more sense, as it is only changing the default if the flag is set. I will try to switch this out today.

brandond:
> I believe the intent was to make an exclusive Plan block any other Plan from running, regardless of any other Plan's settings.

Can we make this configurable as part of controller config? I feel like this is a major change in the Plan contract, that administrators should have to opt into. I think that by default exclusive plans should only block other exclusive plans. If administrators want to allow an exclusive plan to block all other plans, they can enable that in the controller config - with an awareness that this may change the behavior of other plans already deployed to the cluster.

jrodonnell:
> I feel like this is a major change in the Plan contract

In that case, maybe it does make more sense to think of making Plans serial rather than exclusive. The operator could still treat them as exclusive by letting the serial ones finish before submitting more non-serial Plans. The anti-affinity rule would have to be based off a different one than the one here, probably a new unique label (perhaps `upgrade.cattle.io/serial: true`) but I can take a crack at that too.

@davidcassany as the one who opened the original issue I am also curious to hear what you think of this discussion, do you think this new flag should force exclusivity of the Plan with the flag and prevent any other Plan from running or enforce serial scheduling of all Plans with the flag and ones without continue being scheduled as normal now?

brandond:
> The operator could still treat them as exclusive by letting the serial ones finish before submitting more non-serial Plans.

I also want to keep in mind that lots of folks don't treat Plans as a one-shot, and clean them up afterwards. There may be a standing plan to upgrade to the latest release in a channel, for example. In that case the Plan will finish, but execute again automatically when the channel is updated.

pkg/upgrade/job/job.go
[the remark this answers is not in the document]
jrodonnell: Pushed this change but left the flag as-is just for now.

jrodonnell:
> I also want to keep in mind that lots of folks don't treat Plans as a one-shot operations and delete them afterwards. There may be a standing plan to upgrade to the latest release in a channel, for example. In that case the Plan will finish, but execute again automatically when the channel is updated.

That is some good additional context to have, I did not realize that was a common use case. I think it sheds some more light on the original request, which was to provide a way to prevent concurrent Plans from running on a cluster. But long-running or not, I think it's reasonable to expect an operator to tag their Plan appropriately if it is potentially disruptive/sensitive to disruption or not, so I think I agree with what you're saying.

I think this would work: give every Plan Pod a new label, e.g. `upgrade.cattle.io/exclusive: false | true` and if false the current default anti-affinity rule is used, effectively ignoring this new label, and if true then the anti-affinity rule will be based on this new label and any new exclusive Pods will wait their turn.

pkg/apis/upgrade.cattle.io/v1/types.go
jrodonnell: Name changed.

pkg/upgrade/job/job.go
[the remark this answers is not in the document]
jrodonnell: New flag name added into new mutation logic, new label added as well. Tested and working in my own cluster, exclusive Pods will wait for each other (but not for non-exclusive Pods) on a per-node basis and non-`exclusive` ones will still schedule themselves normally.

pkg/apis/upgrade.cattle.io/constants.go
```diff
@@ -15,6 +15,9 @@ const (
 	// LabelController is the name of the upgrade controller.
 	LabelController = GroupName + `/controller`
 
+	// LabelExclusive is the value of the Exclusive flag.
```
brandond, line 18: No tautological doc comments, please - don't just say that it is what it is, but explain what it's for.
```suggestion
	// LabelExclusive is set if the plan should not run concurrent with other plans.
```

pkg/upgrade/job/job.go
```diff
@@ -211,6 +214,21 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
 		*job.Spec.Parallelism = 1
 	}
 
+	if plan.Spec.Exclusive {
+		job.Spec.Template.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution = []corev1.PodAffinityTerm{{
```
brandond, line 218: Curious if we want to replace the default PodAntiAffinity expressions, or just append to them? Appending to them might be more intuitive, and require less boilerplate?

brandond (changes requested):

pkg/upgrade/job/job.go
jrodonnell: I tried that initially, it did not work because the scheduler evaluates multiple anti-affinity rules with a logical OR when making scheduling decisions, so effectively the current scheduling would always happen regardless of additional rules.

brandond: Hmm, that's interesting - the docs say that they should be ANDed:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#podaffinity-v1-core
> **requiredDuringSchedulingIgnoredDuringExecution**
> [PodAffinityTerm](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#podaffinityterm-v1-core) array	If the affinity requirements specified by this field are not met at scheduling time, the pod will not be scheduled onto the node. If the affinity requirements specified by this field cease to be met at some point during pod execution (e.g. due to a pod label update), the system may or may not try to eventually evict the pod from its node. When there are multiple elements, the lists of nodes corresponding to each podAffinityTerm are intersected, i.e. all terms must be satisfied.

Specifically the bit about `all terms must be satisfied` makes it clear that this is logical AND not OR.

jrodonnell: That is very interesting, I did not see that part of the docs when I did my previous testing so I just tested this again and can confirm the docs are not correct, the terms are indeed being OR'd:

1. Made a new commit in my fork using multiple tags [here](https://github.com/rancher/system-upgrade-controller/blob/721d70c57e91036470bc1d2e894e0b4a141069c2/pkg/upgrade/job/job.go#L173)
2. Built & pushed image, deployed to my cluster, applied my test yaml with multiple exclusive Plans
3. Pods all have the correct labels and anti-affinity rules and are still being co-located following the current default scheduling pattern:
```
jrodonnell@k3s0:~$ kubectl get pod -A -o wide --field-selector spec.nodeName=k3s0
NAMESPACE            NAME                                                              READY   STATUS            RESTARTS       AGE    IP            NODE   NOMINATED NODE   READINESS GATES
...
system-upgrade       apply-exclusive-true-1-on-k3s0-with-55d6cc304f29efcbfd4ba-6sxr8   1/1     Running   0              14s    192.168.1.4   k3s0   <none>           <none>
system-upgrade       apply-exclusive-true-2-on-k3s0-with-55d6cc304f29efcbfd4ba-xrcfs   1/1     Running   0              14s    192.168.1.4   k3s0   <none>           <none>

jrodonnell@k3s0:~$ k get pod apply-exclusive-true-1-on-k3s0-with-55d6cc304f29efcbfd4ba-6sxr8 -o json -n system-upgrade | jq -r '.metadata.labels'                                                                                                                                  
{                                                                                                                                                                                                                                                                                  
  "controller-uid": "a62f8de5-88be-4d61-9ea8-4c3878fee9e6",                                                                                                                                                                                                                        
  "job-name": "apply-exclusive-true-1-on-k3s0-with-55d6cc304f29efcbfd4ba-069ef",                                                                                                                                                                                                   
  "plan.upgrade.cattle.io/exclusive-true-1": "55d6cc304f29efcbfd4ba2d6253edb4f91b638b2afe0ce7b99aa8d82",                                                                                                                                                                           
  "upgrade.cattle.io/controller": "system-upgrade-controller",                                                                                                                                                                                                                     
  "upgrade.cattle.io/exclusive": "true",                                                                                                                                                                                                                                           
  "upgrade.cattle.io/node": "k3s0",                                                                                                                                                                                                                                                
  "upgrade.cattle.io/plan": "exclusive-true-1",                                                                                                                                                                                                                                    
  "upgrade.cattle.io/version": "bionic"
}

jrodonnell@k3s0:~$ k get pod apply-exclusive-true-1-on-k3s0-with-55d6cc304f29efcbfd4ba-6sxr8 -o json -n system-upgrade | jq -r '.spec.affinity.podAntiAffinity'
      ...
        "matchExpressions": [
          {
            "key": "upgrade.cattle.io/plan",
            "operator": "In",
            "values": [
              "exclusive-true-1"
            ]
          },
          {
            "key": "upgrade.cattle.io/exclusive",
            "operator": "In",
            "values": [
              "true"
      ...
```

So for our purposes, I think we have to stick with overwriting the value of the rule because we need it to be evaluated, well, exclusively. I think what's happening here is that the code under the hood is in fact ANDing all the terms in MatchExpressions, but for anti-affinity rules it is simply negating the intersection (`~[A && B]`) which by DeMorgan's Law is actually equivalent to `~A || ~B`, hence if one of the rules evaluates to false the Pods are actually scheduled, and only if all the rules evaluate to true is the Pod not scheduled. So for anti-affinity rules, it's effectively only an AND if all the MatchExpressions evaluate to true, otherwise it's OR. I just tested this last case too and that is what I'm seeing, two rules matching Pod labels and none of the Pods are being scheduled so I'm very confident that's what's happening!

brandond (approved):

PR #286, opened 2024-02-12
Propagate labels and annotations from plans to jobs

### LINKED ISSUE #160: Support pod labels and annotations
I need to be able to apply pod annotations for the created jobs so that my log collection framework can route the logs to the indexer.

Support for setting labels and annotations on upgrade.cattle.io/v1:Plan custom resource.

The full path to spec...labels is whatever the owners would want, but it should generate the Job with the .spec.template.metadata. with the labels and annotations provided.

Example implementation: 

```
---
apiVersion: upgrade.cattle.io/v1
kind: Plan
metadata:
  name: job-name
...
spec:
  metadata
    labels:
      my-custom-label: one
    annotations:
      logging.io/index: my-index
  concurrency: 1
...
```

It would then generate job manifests similar to this:

```
apiVersion: batch/v1
kind: Job
metadata:
  name: job-name-hostname
...
spec:
  template:
    metadata:
      name: job-name-hostname1
      labels:
        my-custom-label: one
      annotations:
        logging.io/index: my-index
...
 ```

sfackler:
The `cattle.io` filtering seems a bit sketchy - is there a better way to do that?

Closes #160

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -90,77 +90,6 @@ volumes:
 
 ---
 kind: pipeline
-name: s390x
-
-platform:
-  os: linux
-  arch: amd64
-
-node:
-  arch: s390x
-
-steps:
-- name: build
-  image: rancher/dapper:v0.6.0
-  commands:
-  - dapper ci
-  volumes:
-  - name: docker
-    path: /var/run/docker.sock
-
-- name: upload-artifacts
-  image: rancher/drone-images:github-release-s390x
-  settings:
-    api_key:
-      from_secret: github_token
-    prerelease: true
-    checksum:
-    - sha256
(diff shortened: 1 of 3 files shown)


dweomer (changes requested):
This looks alright to me but some tests that demonstrate it working as expected, for both inclusions and exclusions, are a must.

sfackler:
Updated!

sfackler:
I think the CI may be a bit broken? The s390x and manifest builds seem to never start running.

brandond:
yes, we no longer have s390x runners; those steps and the s390x manifest need to be removed.

sfackler:
Build's green!

dweomer (approved):

brandond (approved):

PR #288, opened 2024-02-17
feat: Reduce permissions for system-upgrade-controller serviceaccount

SISheogorath:
This patch drastically cuts down the permissions of the system-upgrades-controller from the previous cluster-admin permissions to a tailored set of permissions for the controller.

### DIFF
--- a/kustomization.yaml
+++ b/kustomization.yaml
@@ -1,6 +1,8 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
+- manifests/clusterrole.yaml
+- manifests/clusterrolebinding.yaml
 - manifests/system-upgrade-controller.yaml
 images:
 - name: rancher/system-upgrade-controller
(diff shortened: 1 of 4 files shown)


dweomer (approved):
LGTM

SISheogorath:
Before we merge that, I had one more thought: Currently the controller installs its own CRDs if they aren't present in the cluster. So maybe it needs some more permissions for creating CRDs? (I install them separately as part of the kustomization, so it wasn't a thing for me.)

SISheogorath:
If it can't read crds directly, it just assumes they exist and assumes it's all fine.

We could include the CRD itself into the kustomize output.

brandond:
Can you rebase this on top of current master so that CI can finish?

SISheogorath:
There we go :)

dereknola (approved):

brandond (changes requested):
Can you remove the changes to the image/taq kustomization from this PR? I'm not understanding what those have to do with the RBAC; if you want to change that, please open another PR.

SISheogorath:
They had nothing to do with the change, just cleaning up along the way. Undid them for now :)

PR #292, opened 2024-02-27
Add a Complete condition to plans

### LINKED ISSUE #291: Add a Complete condition to Plans when all nodes are up to date
It's currently a bit tricky to figure out if a `Plan` is "complete" - you need to check that every node matching the plan's selector has a `plan.upgrade.cattle.io/my-job` label matching the plan's `latestHash`. While this is doable, I don't think it can be done in a one-liner, and in particular means that you can't `kubectl wait` on a plan completing.

It'd be great if the SUC would add a `Complete: True` condition to the plan when all of the selected nodes have completed the current revision of the plan. This would align with e.g. the k8s `Job`'s conditions and allow you to easily wait on completion: `kubectl wait --for=condition=Complete plan my-plan`.

Some things like rke2-upgrade try to do this by looking at the `status.applying` field, but that seems a bit racy - you might load the plan before it's been mutated by the SUC and think it's complete.

**Additional context**
Unlike a job, a plan can transition out of the completed state if new nodes start to match the plan's selector but I don't think that's too surprising.

sfackler:
It doesn't feel great to be updating the condition in the generating handler instead of the status handler, but this ensures the condition is in sync with the actual state of the plan.

I poked around a bit in the ETE test infrastructure but wasn't totally sure how to add a test for this.

Closes #291

### DIFF
--- a/e2e/suite/job_generate_test.go
+++ b/e2e/suite/job_generate_test.go
@@ -56,6 +56,8 @@ var _ = Describe("Job Generation", func() {
 				WithTimeout(30 * time.Second).
 				Should(WithTransform(upgradeapiv1.PlanComplete.IsTrue, BeFalse()))
 
+			plan, err = e2e.GetPlan(plan.Name, metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
 			plan.Spec.Upgrade.Args = []string{"exit 0"}
 			plan, err = e2e.UpdatePlan(plan)
 			Expect(err).ToNot(HaveOccurred())
--- a/e2e/suite/job_generate_test.go
+++ b/e2e/suite/job_generate_test.go
@@ -53,6 +53,7 @@ var _ = Describe("Job Generation", func() {
 
 			plan, err = e2e.GetPlan(plan.Name, metav1.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
+			Expect(upgradeapiv1.PlanComplete.IsTrue(plan)).To(BeFalse())
 
 			plan.Spec.Upgrade.Args = []string{"exit 0"}
 			plan, err = e2e.UpdatePlan(plan)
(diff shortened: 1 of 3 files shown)


sfackler:
Tests are going to flake here due to the s390x removal until #286 lands and I rebase on top of it.

brandond:
Best way to test would probably be to just find one of the existing tests that waits for the job to complete, and check that the status condition is true when it completes. And check the opposite for a job that is failed.

brandond:
I just merged that other PR, you should be able to rebase.

sfackler:
Ah good call - updated one of the job generation tests.

pkg/upgrade/handle_upgrade.go
```diff
@@ -93,6 +93,8 @@ func (ctl *Controller) handlePlans(ctx context.Context) error {
 				concurrentNodeNames[i] = upgradenode.Hostname(node)
 			}
 			obj.Status.Applying = concurrentNodeNames[:]
+			isComplete := len(concurrentNodeNames) == 0 && obj.Spec.Concurrency > 0
```
sfackler, line 96: The concurrency check here isn't quite accurate, it just ends up causing the plan to never be complete rather than always being complete. Not sure if it's worth retaining?

e2e/suite/job_generate_test.go
```diff
@@ -67,6 +68,10 @@ var _ = Describe("Job Generation", func() {
 			Expect(jobs[0].Status.Succeeded).To(BeNumerically("==", 1))
 			Expect(jobs[0].Status.Active).To(BeNumerically("==", 0))
 			Expect(jobs[0].Status.Failed).To(BeNumerically("==", 0))
+
+			plan, err = e2e.GetPlan(plan.Name, metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(upgradeapiv1.PlanComplete.IsTrue(plan)).To(BeTrue())
```
brandond, line 74: You might need to wrap this with an Eventually?

https://onsi.github.io/ginkgo/#using-speccontext-with-gomegas-eventually

sfackler: Fixed, thanks.

brandond (approved):

dereknola (approved):

brandond:
Just did.

PR #296, opened 2024-03-06
fix: Drop unneeded reorder surpression

SISheogorath:
This patch drops the deprecated flag for output reordering. This causes the output to be ordered properly instead of following the order of the resource list.

References:
https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/sortoptions/#legacy-sorting

### DIFF
--- a/scripts/package-controller
+++ b/scripts/package-controller
@@ -34,6 +34,6 @@ if [ "$ARCH" = "amd64" ]; then
   cp kustomization.yaml /tmp/
   trap reset-kustomization EXIT
   kustomize edit set image "rancher/system-upgrade-controller=${REPO}/system-upgrade-controller:${VERSION}"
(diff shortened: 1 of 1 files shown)


brandond:
lgtm, but I'm curious what the problem is with the current ordering?

brandond (approved):

SISheogorath:
I don't think there is a problem, it's just not the default. This results in the clusterrole, role and clusterrolebinding being created first before the namespace. I think the apiserver figures that out for us already, but with default reordering the Namespace moves to the top and becomes the first object being created.

rpo-fr:
look at this: https://github.com/rancher/system-upgrade-controller/releases/#:~:text=%3A%20v0.13.3...v0.13.4-,Important%20Upgrade%20Notice,-If%20upgrading%20to

PR #297, opened 2024-03-06
Update/fix README kustomize command

SISheogorath:
Before this patch always version v0.8.0 would be deployed when using the
kustomize command from the README. This patch adjust the kustomization
to not overwrite the version.

The build process is unaffected from this, since during the build script
the kustomize file is adjusted to adjust the version to the current
release. This means the build manifests for release will still have a
pinned version like before.

Reference:
https://github.com/rancher/system-upgrade-controller/blob/4eacc2dabbde943ee59dc7819ea0a04909b05b0f/scripts/package-controller#L36

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ But in the time-honored tradition of `curl ${script} | sudo sh -` here is a nice
 
 ```shell script
 # Y.O.L.O.
-kustomize build github.com/rancher/system-upgrade-controller | kubectl apply -f - 
+kubectl apply -k github.com/rancher/system-upgrade-controller
 ```
 
 ### Example Plans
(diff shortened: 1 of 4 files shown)


SISheogorath:
One problem with the command and the new permissions is, that the controller will lack its crds and the permissions to deploy them itself. This is a result of my decision to not grant the controller full permissions to adjust CRDs.

It might be a good idea to revise this or figure out a better way to get the CRDs easily into the cluster.

brandond (approved):

brandond:
I guess I could finally cave to long-standing requests to include the CRD manifest in the repo. I've been trying to keep the CRDs code-first, but since we're suggesting people deploy with kustomize, and the RBAC no longer allows the controller to install them itself, we might as well add it.

SISheogorath:
I think we should add the permissions to create CRDs. The checked in CRDs will age the same way the kustomiztation image replacements will age: poorly.

SISheogorath:
There we go, also added the missing permissions :)

brandond (approved):

Issue #302, opened 2024-03-24
"latest" tag does not exist on docker hub


ShakataGaNai:
Running `kubectl apply -f https://raw.githubusercontent.com/rancher/system-upgrade-controller/master/manifests/system-upgrade-controller.yaml`, as specified in the README.md will fail. That file references `image: rancher/system-upgrade-controller:latest` which does not exist on docker hub today https://hub.docker.com/r/rancher/system-upgrade-controller/tags.

![image](https://github.com/rancher/system-upgrade-controller/assets/1415068/f2653f9a-4cb8-44d9-884e-bef4d654bbf1)

Manual test screenshot.

onno204:
@SISheogorath Could you look into this issue?
It seems to be caused by your commit [4e31e5d6a0926df9f370a06c9ea3250ee1e088b1](https://github.com/rancher/system-upgrade-controller/commit/4e31e5d6a0926df9f370a06c9ea3250ee1e088b1).

This currently breaks some installations because the `latest` tag does not exists on the [docker hub](https://hub.docker.com/r/rancher/system-upgrade-controller/tags). I suggest reverting the commit, and recommiting once the latest tag has been added to the docker hub.

mysticaltech:
Tons of issues for hour project https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner because of this. Please fix ASAP folks.

SISheogorath:
I mean, rolling back is an option, but will also mean you will rollout a 1 year old version of the controller. Of course that's better than not functioning at all, but also not ideal. If we revert the change, we should at least deploy the latest version of the controller.

mysticaltech:
@SISheogorath Was thinking the same. If I understood correctly, I need to do is remove the latest from the manifest via our kustomization and replace with the latest tag on docker hub, please confirm if possible?

mysticaltech:
@SISheogorath Thank you, appreciate it.

mysticaltech:
> The quickest solution would be:
> 
> ```yaml
> # kustomization.yaml
> apiVersion: kustomize.config.k8s.io/v1beta1
> kind: Kustomization
> resources:
> - github.com/rancher/system-upgrade-controller
> images:
> - name: rancher/system-upgrade-controller
>   newTag: v0.13.4
> ```
> 
> This would include all manifests and set the correct version (for now).

Here's the same in kustomization patch format for anyone that needs it, the following goes into the patch array:

```tf
    {
      target = {
[...]

PR #305, opened 2024-03-28
ci(drone): Add push step for latest manifest

SISheogorath:
This patch adds a push step for a latest image. Various CI steps actually build a latest image but it's never pushed. This patch adds a step in the release pipeline to also push the release as latest to Docker Hub.

This should resolve issues people experience when using plain manifests from the repository which still references a latest image, which currently doesn't exist.

Please not: I have no insight into the CI process here, so let me know if this is not working for some reason.

References:
https://github.com/rancher/system-upgrade-controller/issues/302

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -205,6 +205,26 @@ steps:
     - refs/tags/*
     event:
     - tag
+- name: controller-latest
+  image: plugins/manifest
+  settings:
+    username:
+      from_secret: docker_username
+    password:
+      from_secret: docker_password
+    platforms:
+      - linux/amd64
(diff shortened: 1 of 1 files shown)


mysticaltech:
@SISheogorath Will this take time to merge? If so, would it be possible to manually push the "latest" to https://hub.docker.com/r/rancher/system-upgrade-controller/tags. 

Because of this, I've decided to not do the kustomization change. But if it takes time, will do it.

brandond:
no longer using Drone. Also, we tend to not use `latest` for our images, since it is an anti-pattern.

PR #306, opened 2024-04-03
fix imagePullPolicy

simonfelding:
If imagePullPolicy is set to IfNotPresent and the tag is :latest, you get very unintuitive behavior where Kubernetes will not actually get new versions. This kind of behavior is exactly why it's bad practice to use that tag.

Source: https://kubernetes.io/docs/concepts/containers/images/#updating-images

### DIFF
--- a/manifests/system-upgrade-controller.yaml
+++ b/manifests/system-upgrade-controller.yaml
@@ -67,7 +67,7 @@ spec:
       containers:
         - name: system-upgrade-controller
           image: rancher/system-upgrade-controller:latest
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: Always
           securityContext:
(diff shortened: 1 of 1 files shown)


SISheogorath:
It would probably be better to drop the explicit definition of the image pull policy and let the Kubernetes mutation webhook do its magic for the image tags.

brandond:
This is actually updated to point at the release tag as part of publish CI. If you consume the manifest from the release artifacts, instead of just applying what's in git, you should get the correct behavior.

```
brandond@dev01:~$ curl -sL https://github.com/rancher/system-upgrade-controller/releases/download/v0.13.4/system-upgrade-controller.yaml | grep image:
        image: rancher/system-upgrade-controller:v0.13.4
```

PR #310, opened 2024-05-15
chore: Upgrade kubectl to version 1.30.0

SISheogorath:
Just a regular upgrade, since 1.25.4 is a bit dated by now. Also making sure we actually used the docker.io image unless explicitly changed.


IT-Cru:
Does this currently also blocks upgrades to k3s 1.30.x? If yes, would be nice to get it merged :)

SISheogorath:
I don't think it should have that impact, it's just updates for good measure. But you can validate it the impact on your cluster, if you want. All you need to do is adjusting the environment variable as shown in the manifest.

brandond:
kubectl is widely compatible with both newer and older versions of the apiserver. Upgrading or not upgrading doesn't block or unblock anything.

manifests/system-upgrade-controller.yaml
```diff
@@ -22,7 +22,7 @@ data:
   SYSTEM_UPGRADE_JOB_ACTIVE_DEADLINE_SECONDS: "900"
   SYSTEM_UPGRADE_JOB_BACKOFF_LIMIT: "99"
   SYSTEM_UPGRADE_JOB_IMAGE_PULL_POLICY: "Always"
-  SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE: "rancher/kubectl:v1.25.4"
+  SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE: "docker.io/rancher/kubectl:v1.30.0"
```
brandond, line 25:
```suggestion
  SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE: "docker.io/rancher/kubectl:v1.30.4"
```

pkg/upgrade/job/job.go
```diff
@@ -24,7 +24,7 @@ const (
 	defaultBackoffLimit            = int32(2)
 	defaultActiveDeadlineSeconds   = int64(600)
 	defaultPrivileged              = true
-	defaultKubectlImage            = "rancher/kubectl:v1.25.4"
+	defaultKubectlImage            = "docker.io/rancher/kubectl:v1.30.0"
```
brandond, line 27:
```suggestion
	defaultKubectlImage            = "docker.io/rancher/kubectl:v1.30.4"
```

brandond (changes requested):
lgtm, but lets bump this to latest before merging

brandond:
nm, we actually already bumped these in https://github.com/rancher/system-upgrade-controller/pull/326

PR #311, opened 2024-06-06
Add GHA workflows

bfbachmann:
Added GHA workflows:
1. **CI on Pull Request**: Runs tests on arm64 and adm64 on any PR.
2. **CI on Release Tag**: Runs tests, creates a GH release with artifacts, and builds and pushes Docker containers. Example images produced by the build can be found [here](https://hub.docker.com/repository/docker/brunobachmann499/system-upgrade-controller/general). Example releases with artifacts can be found [here](https://github.com/bfbachmann/system-upgrade-controller/releases).
3. **FOSSA Scans**: Just runs security scans on master.

SURE-8004


brandond (approved):

.github/workflows/fossa.yaml
```diff
@@ -0,0 +1,34 @@
+name: FOSSA Scanning
+
+on:
+  push:
+    branches: ["main", "master", "release/**"]
```
jiaqiluo, line 5:
```suggestion
    branches: [ "master" ]
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,227 @@
+name: CI on Release Tag
+
+# The jobs below will execute any time a tag is pushed to any branch in the repo
+# and when master is updated.
+on: 
+  push:
+    branches:
+      - 'master'
```
jiaqiluo, line 8: I think it will result in the creation of a tag named `master`. 

If we check [this Drone run](https://drone-publish.rancher.io/rancher/system-upgrade-controller/214) in the event of pushing to master, only the steps of clone and build are expected. 

I suggest moving this condition ( on.push.branches.master) to the file `.github/workflows/pull-request.yaml`, and modifying that filename and CI name according.

jiaqiluo (changes requested):
Left some comments. 
Also, I think we want to remove the `.drone.yaml` file as a part of this PR.

jiaqiluo (approved):

Issue #294, opened 2024-02-29
Failed to upgrade from v0.13.2 to v0.13.3


aofei:
**Version**

v0.13.2 -> v0.13.3

**Platform/Architecture**

Ubuntu 22.04, amd64

**Describe the bug**

```
clusterrole.rbac.authorization.k8s.io/system-upgrade-controller created
role.rbac.authorization.k8s.io/system-upgrade-controller created
clusterrole.rbac.authorization.k8s.io/system-upgrade-controller-drainer created
rolebinding.rbac.authorization.k8s.io/system-upgrade created
namespace/system-upgrade unchanged
serviceaccount/system-upgrade unchanged
configmap/default-controller-env unchanged
deployment.apps/system-upgrade-controller configured
Error from server (Invalid): error when creating "https://github.com/rancher/system-upgrade-controller/releases/download/v0.13.3/system-upgrade-controller.yaml": ClusterRoleBinding.rbac.authorization.k8s.io "system-upgrade-drainer" is invalid: subjects[0].namespace: Required value
Error from server (Invalid): error when applying patch:
{"metadata":{"annotations":{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"rbac.authorization.k8s.io/v1\",\"kind\":\"ClusterRoleBinding\",\"metadata\":{\"annotations\":{},\"name\":\"system-upgrade\"},\"roleRef\":{\"apiGroup\":\"rbac.authorization.k8s.io\",\"kind\":\"ClusterRole\",\"name\":\"system-upgrade-controller\"},\"subjects\":[{\"kind\":\"ServiceAccount\",\"name\":\"system-upgrade\"}]}\n"}},"roleRef":{"name":"system-upgrade-controller"},"subjects":[{"kind":"ServiceAccount","name":"system-upgrade"}]}
to:
Resource: "rbac.authorization.k8s.io/v1, Resource=clusterrolebindings", GroupVersionKind: "rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding"
Name: "system-upgrade", Namespace: ""
[...]

PR #313, opened 2024-06-26
Fix clusterrolebinding by adding a namespace to sa

manuelbuil:
In custerrolebindings, serviceaccounts need to have a namespace, otherwise we get the error:
```
ClusterRoleBinding.rbac.authorization.k8s.io "system-upgrade-drainer" is invalid: subjects[0].namespace: Required value
```
This PR fixes that by adding the namespace `system-upgrade`, which is what that serviceAccount has: https://github.com/rancher/system-upgrade-controller/blob/v0.13.4/manifests/system-upgrade-controller.yaml#L12

### DIFF
--- a/manifests/clusterrolebinding.yaml
+++ b/manifests/clusterrolebinding.yaml
@@ -10,6 +10,7 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: system-upgrade
+  namespace: system-upgrade
 ---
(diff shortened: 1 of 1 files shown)


brandond:
See the discussion at https://github.com/rancher/system-upgrade-controller/issues/294#issuecomment-1970602865 - the lack of NS is intentional.

manuelbuil:
Thanks @brandond for the heads up

Issue #63, opened 2020-04-16
Provide optional upgrade window setting to limit when a plan can run


bashofmann:
**Is your feature request related to a problem? Please describe.**
It would be great if you could limit the execution of a plan to a specific timeframe, so that the upgrade jobs of a plan can for example only run:
* between 2am and 4am
* every Wednesday between 2pm and 6pm
* ...

**Describe the solution you'd like**
I was thinking about adding to configurations `upgrade-window-start` and `upgrade-window-length` that work similar to the reboot windows in the container-linux-update-operator (https://github.com/coreos/container-linux-update-operator/blob/master/doc/reboot-windows.md).
The configuration flags could be optionally added as annotations to a Node and the system-upgrade-controller would only schedule a Job on this node if the current time is currently within this timeframe.
By adding the configuration to the Node one could easily configure different upgrade windows on different set of nodes within a cluster. A node setting would affect all plans the same way.

**Describe alternatives you've considered**
Another option would be to add the configuration flags globally to the system-upgrade-controller as environment variables and/or CLI arguments, but this would decrease the flexibility.

In addition the configuration flag could als be set as a field in the spec of a plan, but I'd propose to discuss this independently, because it adds more complexity to the feature:
* What happens when the upgrade window of a plan conflicts with the upgrade window of Nodes? The plan would never run them.
* How would an upgrade window setting of a plan interact with a use case where you want to schedule plans in a cron like way? Is this even something that should be supported or is the channel feature a better solution for this use-case?

Because of this I feel starting with adding these settings to Nodes is a good first step, that would already solve a large amount of use-cases.

**Additional context**
Once we decide on an implementation, I'll be happy to implement it and create a PR for this.

mstrent:
Defining maintenance windows would be a very useful feature for us as well.

ddfreiling:
We could really use this. Upgrading during peak hours or weekends is not ideal for us.

tuxpeople:
Would like to have that feature as well. I'm about to pitch the use of system-upgrade-controller for a larger project with a few hundred small (one-node) edge clusters. As they are one-node clusters, they should not upgrade during peak hours.

I propose something like Kured has: https://kured.dev/docs/configuration/#setting-a-schedule with those four options, most of the use-cases can be addressed IMHO.

brandond:
It looks like kured did a nice job of packaging up their `timewindow` code in a reusable package, we could probably just pull that in directly.

[...]

PR #316, opened 2024-07-03
Add Optional Upgrade Window

justdan96:
This PR is to add a set of new optional parameters. These allow a user to specify an upgrade window, and plans will fail to validate until they are inside that upgrade window. This allows us to create a plan and then wait for the controller to pick it up during the window.

A lot of this code was inspired by (or in the case of `pkg/upgrade/periodic.go` and `insideUpgradeWindow()`, copied from) https://github.com/flatcar/flatcar-linux-update-operator. As both projects have the same licence I believe this should be compatible.

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ var (
 	debug                               bool
 	kubeConfig, masterURL               string
 	namespace, name, serviceAccountName string
+	windowStart, windowLength           string
 	threads                             int
 )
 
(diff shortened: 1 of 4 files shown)


brandond:
In the [issue](https://github.com/rancher/system-upgrade-controller/issues/63) tracking this feature, there was discussion of using the timewindow package from kured. Any reason you chose to copy code from flatcar instead of using a library>

brandond:
I'm also curious why anyone would want this set at a controller level, instead of per-plan.

justdan96:
I must have missed that issue when looking through, or I assumed nobody was looking at it due to the age. If you are developing that I will close this PR down. 

If you have a regular maintenance window whether the window is defined in the controller or the plan the end result is the same surely?

brandond:
Well, if you make it controller-level then anyone who wanted different windows for different plans is out of luck - they'd have to run multiple upgrade controller deployments just to be able to upgrade different things at different times.

I could see allowing for a **default** upgrade window, in addition to allowing individual plans to have their own window, or opt out of a window entirely if they can run whenever - but only having a single system-wide window does not seem very flexible, and the SUC is intended to be all about flexibility.

I don't believe we have anyone working on it at the moment, but if you wanted to update this PR to function as discussed, that would definitely be appreciated!

Issue #314, opened 2024-07-02
When to expect a new release?


harsimranmaan:
There are some public golang CVEs that requires addressing in the suc. Per the security policy, these are patched during the dev cycle. Is there a cadence to expect such patch releases?

brandond:
There are a couple packaging issues I want to fix before we do another release. It is on my radar for sometime in the next week or two.

buroa:
@brandond Any updates?

harsimranmaan:
Hi team, it would great if a new release could be published as suc gets flagged for multiple critical vulns. The patches have been merged already,

kashalls:
@brandond Do you need any help to get this moving?

brandond:
Sorry, there was a bunch of release CI stuff to fix - the changes from https://github.com/rancher/system-upgrade-controller/pull/311 did not actually work to move publish CI over to GHA.

[v0.14.0](https://github.com/rancher/system-upgrade-controller/releases/tag/v0.14.0) should work.

harsimranmaan:
Thanks Brandon but could the release be bumped to v0.15.0? 0.14.0 was likely published in the past and recalled it seems as gosumdb already has entries for it with a different shasum. Please see https://pkg.go.dev/github.com/rancher/system-upgrade-controller?tab=versions

brandond:
I'm not able to find any references to that tag on GH or Docker Hub, so I have no idea where that would have come from. I can tag 0.15.0 next week when I am back in the office.

PR #326, opened 2024-08-08
Upgrade outdated golang dependencies

harsimranmaan:
Some Golang CVEs get flagged in the suc binary as it was built with an older golang version. Use go 1.22.6
Updating other go module deps.
SLES update is required for libc compatiblity with latest ginkgo version.

Update golangci-lint. Most gingko and gomega changes are to fix linting failures.
Update sonobuoy

docker bridge address is added to tls-san to avoid cert errors when using the client to connect the KUBEHOST.

Drop deprecated version field in the docker compose manifests.

It is recommended that the next release be v0.15.0 as 0.14.0 was already published and deleted in the past and gosumdb already has entries.

Tested manually with a drop in replacement for the older suc binary for k3s upgrade from v1.28.7 to 1.29.6. 

Passing test suite: https://github.com/harsimranmaan/system-upgrade-controller/pull/1

Relates #314


e2e/suite/channel_resolve_test.go
```diff
@@ -5,17 +5,17 @@ import (
...
-	. "github.com/onsi/gomega"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
 	"github.com/rancher/system-upgrade-controller/e2e/framework"
 	upgradeapiv1 "github.com/rancher/system-upgrade-controller/pkg/apis/upgrade.cattle.io/v1"
 	upgradeplan "github.com/rancher/system-upgrade-controller/pkg/upgrade/plan"
 )
 
-var _ = Describe("Resolve channel", func() {
+var _ = ginkgo.Describe("Resolve channel", func() {
```
brandond, line 15: our pattern in is to import the DSL into the package namespace - not to import it as ginkgo/gomega. These changes don't appear to be related to package updates, please drop these from the PR.

brandond (changes requested):
Thanks for the updates! one nit on the test framework changes.

e2e/suite/job_generate_test.go
```diff
@@ -6,12 +6,12 @@ import (
 	"io"
 	"time"
 
+	. "github.com/onsi/ginkgo/v2" //nolint:revive
```
brandond, line 9: Can you perhaps disable whatever check is failing in the `.golangci.json` instead of adding nolint comments to the imports?

brandond (approved):

PR #348, opened 2025-02-17
Move helm charts from rancher/charts to /charts dir

pratikjagrut:
### PR Description  

#### **Why This Change?**  

#### **What’s Changing?**  
- A new branch, `rancher-charts-release/v2.10`, will be used exclusively for SUC charts.  


brandond (changes requested):
What exactly is going on here? You appear to have deleted almost everything from the repo, and replaced it with something else entirely.

Did you mean to open this PR against a different repo entirely? I'm not sure exactly how you got here but I don't think you're in the right place.

snasovich:
> What exactly is going on here? You appear to have deleted almost everything from the repo, and replaced it with something else entirely.
> 
> Did you mean to open this PR against a different repo entirely? I'm not sure exactly how you got here but I don't think you're in the right place.

@brandond , this change is specifically to adopt "Auto Chart Bumps" feature as specified in https://github.com/rancher/rancher/issues/49087. Note this PR is against newly created `rancher-charts-release/v2.10` and not `master`. Master is staying intact.

@pratikjagrut , please add relevant info to PR description.

brandond:
rancher/rancher#49087 says 
> Create a new Git repository for system-upgrade-controller, as it is currently included within rancher/charts

So I would expect to see the rancher-specific chart and rancher specific chart branches in a new repo, not here. Why would we want to make this repo more closely tied to Rancher support lifecycle? It is extensively used by standalone customers. If there is a rancher chart that needs rancher branches that should go in a new repo as suggested in the linked issue.

pratikjagrut:
Hi @brandond, does the updated description help answer your question?

brandond:
As per the other issue this was to be done in a dedicated repo. It still feels wrong to me that we are adding new rancher-specific charts and branching strategy to this repo.

snasovich:
Closing the PR to re-create the branch.

Issue #346, opened 2025-02-15
Generated job name is invalid


tuxpeople:
**Version**
v0.15.0-rc2

**Platform/Architecture**
Linux talos-test04 6.12.11-talos #1 SMP Tue Jan 28 09:32:23 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

**Describe the bug**
Jobs generated have invalid names, as they end with a "-":

```
$ kubectl logs -n kube-system system-upgrade-8445f958db-knnpd
[...]
I0215 06:38:31.911115       1 event.go:389] "Event occurred" object="kube-system/talos" fieldPath="" kind="Plan" apiVersion="upgrade.cattle.io/v1" type="Normal" reason="SyncJob" message="Jobs synced for version v1.9.4 on Nodes talos-test02. Hash: "
time="2025-02-15T06:38:31Z" level=error msg="error syncing 'kube-system/talos': handler system-upgrade: secrets \"system-upgrade\" not found, handler system-upgrade: failed to create kube-system/apply-talos-on-talos-test02-with- batch/v1, Kind=Job for system-upgrade kube-system/talos: Job.batch \"apply-talos-on-talos-test02-with-\" is invalid: [metadata.name: Invalid value: \"apply-talos-on-talos-test02-with-\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.template.labels: Invalid value: \"apply-talos-on-talos-test02-with-\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')], requeuing"
```

**To Reproduce**
1. Deploy v0.15.0-rc2
2. add Plan

Deployment-YAML:
<details>

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "3"
    meta.helm.sh/release-name: system-upgrade
    meta.helm.sh/release-namespace: kube-system
  creationTimestamp: "2025-02-14T12:19:07Z"
[...]

PR #351, opened 2025-03-03
Make IgnoreUpdate secrets optional

brandond:
Fixes issue where nonexistent secret with ignoreUpdate set would cause the plan to endlessly requeue due to errors. If it is ignored for purposes of plan hash, it should be optional.

* https://github.com/rancher/system-upgrade-controller/issues/346

### DIFF
--- a/pkg/upgrade/plan/plan.go
+++ b/pkg/upgrade/plan/plan.go
@@ -79,15 +79,17 @@ func DigestStatus(plan *upgradeapiv1.Plan, secretCache corectlv1.SecretCache) (u
 		}
 
 		for _, s := range plan.Spec.Secrets {
-			secret, err := secretCache.Get(plan.Namespace, s.Name)
-			if err != nil {
-				return plan.Status, err
-			}
 			if !s.IgnoreUpdates {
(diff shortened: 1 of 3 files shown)


pkg/upgrade/plan/plan.go
```diff
@@ -239,7 +241,7 @@ func sha256sum(s ...string) string {
 }
 
 // Validate performs validation of the plan spec, raising errors for any conflicting or invalid settings.
-func Validate(plan *upgradeapiv1.Plan) error {
+func Validate(plan *upgradeapiv1.Plan, secretCache corectlv1.SecretCache) error {
```
VestigeJ, line 244: is this a nit on correct (sp?)

brandond: nit on the spelling of what, the package? That is how its always been imported in this file:
https://github.com/rancher/system-upgrade-controller/blob/9de913246abe4e1d6fb1f04939c5b805a436a44b/pkg/upgrade/plan/plan.go#L20

VestigeJ (approved):
LGTM

PR #352, opened 2025-03-04
support k8s 1.32.2

jiaqiluo:
https://github.com/rancher/rancher/issues/48825

Changes:

1. Support k8s 1.32 

Bump the k8s modules to 1.32.2 
Bump the kubectl

2. Fix the broken e2e test cases  

 

Failed tests:
```
 Job Generation when fails because of invalid time window should apply successfully after edit
 Job Generation when fails because of invalid post complete delay should apply successfully after edit
```

The `initContainer` is only set when `plan.Spec.Prepare`, `plan.Spec.Cordon`, or `plan.Spec.Drain` is not nil [code](https://github.com/rancher/system-upgrade-controller/blob/master/pkg/upgrade/job/job.go#L309-L391), which doesn't seem to be the case for the plan created in these tests.


brandond (approved):

brandond:
Seems like for some reason e2e was not running. Logs for previous PRs say:
>```
>Plugin: system-upgrade-controller
>Status: passed
>Total: 1
>Passed: 1
>Failed: 0
>Skipped: 0
>```

For this PR we have:
>```
>Plugin: system-upgrade-controller
>Status: passed
>Total: 16
>Passed: 16
>Failed: 0
>Skipped: 0
>```

jiaqiluo:
@brandond,  thanks for sharing the findings, that at least explains why the CI passed in the past. 
I have the permission to merge PR and will do it once I get another approval from my team.

PR #356, opened 2025-03-18
[release-v2.10] Move helm charts from rancher/charts to /charts dir

pratikjagrut:
### PR Description  

#### **Why This Change?**  

#### **What’s Changing?**  
- A new branch, `rancher-charts-release/v2.10`, will be used exclusively for SUC charts.  

### DIFF
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -1,0 +1,17 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.25.0-0 < 1.32.0-0'
+  catalog.cattle.io/managed: "true"
+  catalog.cattle.io/namespace: cattle-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/permits-os: linux,windows
+  catalog.cattle.io/rancher-version: '>= 2.10.0-0'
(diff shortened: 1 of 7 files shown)


charts/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.25.0-0 < 1.32.0-0'
+  catalog.cattle.io/managed: "true"
+  catalog.cattle.io/namespace: cattle-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/permits-os: linux,windows
+  catalog.cattle.io/rancher-version: '>= 2.10.0-0'
+  catalog.cattle.io/release-name: system-upgrade-controller
+apiVersion: v1
+appVersion: v0.14.2
```
brandond, line 12: v0.14.2 is from October, can we use something newer?

snasovich: This matches the version used in v2.10.x release line (https://github.com/rancher/charts/blob/dev-v2.10/charts/system-upgrade-controller/105.0.1/values.yaml#L8) so we're just matching that.
Bumping SUC version is out of scope for this effort.

brandond (approved):

snasovich:
This has been deprioritized, closing for now.

PR #357, opened 2025-03-19
[release-v2.11] Move helm charts from rancher/charts to /charts dir

pratikjagrut:
### PR Description  

#### **Why This Change?**  

#### **What’s Changing?**  
- A new branch, `rancher-charts-release/v2.11`, will be used exclusively for SUC charts.  

### DIFF
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -1,0 +1,17 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.25.0-0 < 1.33.0-0'
+  catalog.cattle.io/managed: "true"
+  catalog.cattle.io/namespace: cattle-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/permits-os: linux,windows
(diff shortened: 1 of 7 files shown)


snasovich (changes requested):
There is now https://github.com/rancher/charts/tree/dev-v2.11/charts/system-upgrade-controller/106.0.0 for `v2.11.0` so we should use it as the base.

pratikjagrut:
> There is now https://github.com/rancher/charts/tree/dev-v2.11/charts/system-upgrade-controller/106.0.0 for `v2.11.0` so we should use it as the base.

@snasovich Done.

brandond (approved):

jiaqiluo (approved):
Changes look good to me, they are identical to the chart at https://github.com/rancher/charts/tree/dev-v2.11/charts/system-upgrade-controller/106.0.0 

Hi @pratikjagrut, is it true that, instead of maintaining the (most-of-the-time-idential) templates in multiple branches in the rancher/charts repo, we will maintain them in multiple branches here? So, if we need to add a change to the chart and release a new version of the chart in both Rancher 2.10 and 2.11 release lines, we will need to raise 2 PRs in this repo to the corresponding branches, is it right?

Issue #363, opened 2025-04-16
Prepare securityContext is ignored


maltelehmann:
**Version**
`v0.15.2`

**Platform/Architecture**
`linux/amd64`

**Describe the bug**
When specifying `spec.prepare.securityContext` for a `Plan` CR, the value is not appllied to upgrade Job prepare `initContainer`.

**To Reproduce**
Specify the value and schedule an update.

**Expected behavior**
The `securityContext` of the `prepare` configuration should be used for prepare `initContainer`.

**Actual behavior**
The `securityContext` is ignored.

**Additional context**
I will provide an MR.

PR #364, opened 2025-04-16
Fix securityContext and volumes for prepare

maltelehmann:
fixes https://github.com/rancher/system-upgrade-controller/issues/363


brandond:
Thanks!

Issue #372, opened 2025-08-07
Add support for priorityClassName of Jobs


maltelehmann:
**Is your feature request related to a problem? Please describe.**
We encountered the problems that an update job could not be scheduled due to resource limitations of one host. We had to manually intervene and remove one pod from the host for the job to start. The manual intervention could have been avoided by pod eviction, if the job would have a (high) priority class.

**Describe the solution you'd like**
Adding an optional field `priorityClassName` to the `planSpec` that sets the priority class name for the jobs.

**Describe alternatives you've considered**
Setting a built-in priority class like `system-cluster-critical`. However, this would be a breaking change and might not be the desired behavior for some users. Hence, it is better to make it optional.

**Additional context**
n/a

maltelehmann:
I will provide a PR for this

PR #373, opened 2025-08-09
feat: add optional priority class name to plan spec

maltelehmann:
fixes https://github.com/rancher/system-upgrade-controller/issues/372

### DIFF
--- a/doc/plan.md
+++ b/doc/plan.md
@@ -123,7 +123,7 @@ _Appears in:_
 | `drain` _[DrainSpec](#drainspec)_ | Configuration for draining nodes prior to upgrade. If left unspecified, no drain will be performed. |  |  |
 | `imagePullSecrets` _[LocalObjectReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#localobjectreference-v1-core) array_ | Image Pull Secrets, used to pull images for the Job. |  |  |
 | `postCompleteDelay` _[Duration](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#duration-v1-meta)_ | Time after a Job for one Node is complete before a new Job will be created for the next Node. |  |  |
-
+| `PriorityClassName` _string_ | Priority Class Name of Job, if specified. |  |  |
 
 #### PlanStatus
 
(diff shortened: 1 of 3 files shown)


doc/plan.md
```diff
@@ -123,7 +123,7 @@ _Appears in:_
 | `drain` _[DrainSpec](#drainspec)_ | Configuration for draining nodes prior to upgrade. If left unspecified, no drain will be performed. |  |  |
 | `imagePullSecrets` _[LocalObjectReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#localobjectreference-v1-core) array_ | Image Pull Secrets, used to pull images for the Job. |  |  |
 | `postCompleteDelay` _[Duration](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#duration-v1-meta)_ | Time after a Job for one Node is complete before a new Job will be created for the next Node. |  |  |
-
+| `PriorityClassName` _string_ | Priority Class Name of Job, if specified. |  |  |
```
brandond, line 126: Run `go generate` and check in the resulting change. This should be generated from the doc comment on the type, and does not match what you put there.

brandond (changes requested):

doc/plan.md
maltelehmann: Oh, thank you for the feedback. I run the generation and checked-all changes.

pkg/apis/upgrade.cattle.io/v1/types.go
```diff
@@ -82,6 +82,8 @@ type PlanSpec struct {
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 	// Time after a Job for one Node is complete before a new Job will be created for the next Node.
 	PostCompleteDelay *metav1.Duration `json:"postCompleteDelay,omitempty"`
+	// Priority Class Name of Job, if specified.
+	PriorityClassName *string `json:"priorityClassName,omitempty"`
```
brandond, line 86: Question: why `*string` instead of just `string`? The field in pod spec is just a string. Making this a string would simplify both the generated deepcopy and the assignment in pod spec.

brandond (approved):

PR #337, opened 2024-11-12
Add support for setting secret mount mode

solidDoWant:
This will allow setting `defaultMode` on secret mounts. This should be backwards compatible with existing resources.

PR #374, opened 2025-08-11
Add support for setting secret mount mode

solidDoWant:
Rebase of https://github.com/rancher/system-upgrade-controller/pull/337. This will allow setting `defaultMode` on secret mounts. This should be backwards compatible with existing resources.


pkg/crds/yaml/generated/upgrade.cattle.io_plans.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.17.3
+    controller-gen.kubebuilder.io/version: v0.18.0
```
brandond, line 6: looks like you're using a different version of kube-builder than our CI. Would you mind reverting this bit?

brandond (approved):

PR #382, opened 2025-09-27
bump go version to 1.25

farazkhawaja:
Upgrades go version to 1.25
ran go mod tidy on suc/ and suc/pkg/apis folder


go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/rancher/system-upgrade-controller
 
-go 1.24.6
+go 1.25.0
```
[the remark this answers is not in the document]
hammadsaedi: fyi; BCI got `go 1.25.3`

```
hammad.saeedi@Hammads-MacBook-Pro % docker run registry.suse.com/bci/golang:1.25 go version
go version go1.25.3 linux/arm64
```

farazkhawaja:
@brandond

brandond (changes requested):
Thanks for sticking with it!

```
Running: golangci-lint
Error: can't load config: the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.1)
Failed executing command with error: can't load config: the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.1)
time="2025-11-06T19:42:01Z" level=fatal msg="exit status 3"
```

need to bump golangci-lint as well, for golang 1.25 compat.

farazkhawaja:
@brandond we're using v1.64.8 and thats the latest 1.x.x version released by golangci. moving to 2.x.x would mean a major version bump. Do we wanna proceed with that?

brandond:
> `Error: can't load config: unsupported version of the configuration: "" See https://golangci-lint.run/docs/product/migration-guide for migration instructions`

You can't just bump the golangci-lint version, you need to migrate the config from v1 -> v2 as well :/

.golangci.bck.json
brandond: delete this. This is git, we don't need to keep "backup" files around.

brandond (changes requested):
one change. lgtm otherwise!

.golangci.bck.json
farazkhawaja: done

PR #386, opened 2025-11-14
Extend Upgrade Controller for Windows Node Compatibility

ponmuthudev:
**Summary**
Adds support for Windows nodes by enabling the controller to detect when a node is running Windows and use a Windows-compatible kubectl image and appropriate security context settings.

**Key Changes**
Introduces a new environment variable SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE_WINDOWS for specifying the windows kubectl image to use when performing upgrades targeting Windows nodes. 

**In the job creation logic (in job.go):**
- Detects whether the node OS label (kubernetes.io/os) equals “windows” and switches to the Windows image accordingly. 
- For Windows nodes, sets SecurityContext.WindowsOptions.HostProcess = true and RunAsUserName = "NT AUTHORITY\\SYSTEM" on init-containers (prepare, drain, cordon) and the main upgrade container path. 
- Updates the default manifest (manifests/system-upgrade-controller.yaml) to include the new variable comment and default value for SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE_WINDOWS. 

**Why this is needed**
Up to now the upgrade controller assumed Linux nodes and used a Linux kubectl image and Linux-style security context. Windows nodes have different requirements (e.g., HostProcess containers and Windows security context) which this change enables. Supporting Windows nodes helps extend the controller to heterogeneous clusters.

**Impact & Considerations**
- Operators upgrading Windows worker or master nodes will need to set the SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE_WINDOWS env var to a Windows-e.g., tagged kubectl image.
- The logic enforces a fatal error if the env var is not set when targeting a Windows node.
- Security contexts differ: Windows containers require HostProcess=true and run under NT AUTHORITY\SYSTEM, so this may affect privileges and cluster security; review accordingly.
- Existing Linux-only flows remain unchanged, so backward compatibility is maintained.

**Testing & Validation**
- Verified that for a node labelled kubernetes.io/os=windows, the Windows image is picked and the pod spec has the Windows security settings.
- Verified Linux nodes continue to use the original image and security context path.
- Verified that the upgrade workflow successfully triggers on both Linux and Windows nodes.

### DIFF
--- a/manifests/system-upgrade-controller.yaml
+++ b/manifests/system-upgrade-controller.yaml
@@ -24,11 +24,11 @@ data:
   SYSTEM_UPGRADE_JOB_BACKOFF_LIMIT: "99"
   SYSTEM_UPGRADE_JOB_IMAGE_PULL_POLICY: "Always"
   SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE: "rancher/kubectl:v1.30.3"
+  # Only set if you have Windows nodes to upgrade; ignored otherwise.
+  SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE_WINDOWS: ""
   SYSTEM_UPGRADE_JOB_PRIVILEGED: "true"
   SYSTEM_UPGRADE_JOB_TTL_SECONDS_AFTER_FINISH: "900"
   SYSTEM_UPGRADE_PLAN_POLLING_INTERVAL: "15m"
-  # Only set if you have Windows nodes to upgrade; ignored otherwise.
-  SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE_WINDOWS: ""
 ---
 apiVersion: apps/v1
 kind: Deployment


pkg/upgrade/job/job.go
```diff
@@ -310,18 +311,42 @@ func New(plan *upgradeapiv1.Plan, node *corev1.Node, controllerName string) *bat
...
+
+	// Initialize with the default kubectl image; can be changed to Windows or other image if needed
+	selectedKubectlImage := KubectlImage
+	if isWindows {
+		// Retrieve the Windows kubectl image from the SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE_WINDOWS environment variable.
+		// This environment variable must be set by customers/operators when performing upgrades on Windows nodes.
+		selectedKubectlImage = os.Getenv("SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE_WINDOWS")
+		if selectedKubectlImage == "" {
+			logrus.Fatal("SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE_WINDOWS is a required environment variable when targeting Windows")
+		}
```
brandond: Killing the whole controller over this seems like an overreaction. Other plans that don't run against Windows nodes will still work fine, as will any nodes in THIS plan that are not Windows.

Do this env var lookup in the same place we do all the OTHER env var lookups for global settings, and then just fail plan validation if any of the nodes that match the nodeSelector are windows but this variable is not set.

PR #449, opened 2026-08-28
Bump k8v1.37

Abhishek-Valaboju:
https://github.com/rancher/rancher/issues/55781
add support for kubernetes v1.37


go.mod
```diff
@@ -7,146 +7,167 @@ toolchain go1.26.5
...
+	k8s.io/externaljwt => k8s.io/externaljwt v0.37.0
+	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.37.0
+	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.37.0
+	k8s.io/kube-proxy => k8s.io/kube-proxy v0.37.0
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.37.0
+	k8s.io/kubelet => k8s.io/kubelet v0.37.0
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.37.0
+	k8s.io/metrics => k8s.io/metrics v0.37.0
+	k8s.io/mount-utils => k8s.io/mount-utils v0.37.0
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.37.0
```
brandond: nit: I think we can drop these replaces and just require the correct versions now? No need to bump the versions twice (replace+require).

brandond (approved):

jakefhyde (approved):

go.mod
Abhishek-Valaboju: Just to confirm, should we drop the replace only for the entries that are already duplicated in require, or the full k8s.io block in replace, including the ones with no require entry at all?

brandond: Drop all the replacements, for both direct and indirect k8s.io deps. Replacing something with the same version that is required does not serve any purpose.

Abhishek-Valaboju: Done

brandond (approved):

